PR #149, opened 2022-02-23
Introduce transactions per address table

tdroxler:
I still want to clean few stuff, but it's good to start to review as it's a big topic.

Sorry the commits aren't super well split, but I tried to make them more digestible. 

I synced locally and compared with prod, we have the exact same data, it's faster but it's a bit unfair as I have a good machine. But based on the benchmarks queries should be 2-3 times faster.
maim-1645603143
That's the gate.io address, which has the most transactions.


tdroxler:
Please note that I haven't added yet the `address` field inside the Input table, as it's not linked to getting the transactions' hash for a given address, I preferred not to put too much on that PR

app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
```diff
@@ -217,5 +218,11 @@ object BlockDao {
         )
       }
     }
+
+    def updateInputs(inputs: Seq[InputEntity]): Future[Int] = {
+      run(
+        DBIOAction.sequence(inputs.map(insertTxPerAddressFromInput))
+      ).map(_.sum)
+    }
```
simerplaha, line 226: Would there a need for these inserts to be within a transaction? My guess is maybe not because in case of partial success the failed inserts would get retried, is that correct?

simerplaha (approved):
Looks great! Just added one question for my benefit 🚀  🚀

app/src/main/scala/org/alephium/explorer/persistence/model/TransactionPerAddressEntity.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.alephium.explorer.api.model.{Address, BlockEntry, Transaction}
+import org.alephium.util.TimeStamp
+
+final case class TransactionPerAddressEntity(
+    hash: Transaction.Hash,
+    blockHash: BlockEntry.Hash,
+    timestamp: TimeStamp,
+    txIndex: Int,
+    address: Address,
+    mainChain: Boolean
```
polarker, line 28: Shall we put `address` as the first field of the class ?

app/src/main/scala/org/alephium/explorer/persistence/queries/InputsQueries.scala
```diff
@@ -0,0 +1,165 @@
...
+  val config: DatabaseConfig[JdbcProfile]
+  import config.profile.api._
+
+  private val mainInputs  = inputsTable.filter(_.mainChain)
+  private val mainOutputs = outputsTable.filter(_.mainChain)
+
+  def insertTxPerAddressFromInputs(inputs: Seq[InputEntity],
+                                   outputs: Seq[OutputEntity]): DBActionW[Seq[InputEntity]] = {
+
+    val inputsWithMaybeAddress = inputs.map(in => (in.address, in))
```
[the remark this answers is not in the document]
polarker: there is no need to create `inputsWithMaybeAddress` object AFAIS

app/src/main/scala/org/alephium/explorer/persistence/queries/InputsQueries.scala
```diff
@@ -0,0 +1,165 @@
...
+
+    val assets = inputsWithMaybeAddress
+      .collect {
+        case (Some(address), input)
+            if !outputs.exists(out => out.address == address && out.txHash == input.txHash) =>
```
polarker, line 50: This would be insanely slow when there are many inputs and outputs in the one block. The complexity is O(inputs.length * outputs.length).

app/src/main/scala/org/alephium/explorer/persistence/queries/InputsQueries.scala
```diff
@@ -0,0 +1,165 @@
...
+          case (address, input) =>
+            val instant = java.time.Instant.ofEpochMilli(input.timestamp.millis)
+            s"('\\x${input.txHash}', '\\x${input.blockHash}', '${instant}', ${input.txIndex}, '${address}', ${input.mainChain}) "
+        }
+        .mkString(",\n")
+
+      sqlu"""
+      INSERT INTO transaction_per_addresses (hash, block_hash, timestamp, tx_index, address, main_chain)
+      VALUES #$values
+      ON CONFLICT (hash, block_hash, address) DO NOTHING
```
polarker, line 97: There should be no conflict. We should throw an error in this case

app/src/main/scala/org/alephium/explorer/persistence/queries/OutputsQueries.scala
```diff
@@ -0,0 +1,119 @@
...
+      val values = outputs
+        .map { output =>
+          val instant = java.time.Instant.ofEpochMilli(output.timestamp.millis)
+          s"('\\x${output.txHash}','\\x${output.blockHash}','${instant}', ${output.txIndex},'${output.address}',${output.mainChain})"
+        }
+        .mkString(",\n")
+      sqlu"""
+      INSERT INTO transaction_per_addresses (hash, block_hash, timestamp, tx_index, address, main_chain)
+      VALUES #$values
+      ON CONFLICT (hash, block_hash, address) DO NOTHING
```
polarker, line 53: same here

polarker (commented):
Done my first round of review. We need to improve the design a bit more

app/src/main/scala/org/alephium/explorer/persistence/queries/InputsQueries.scala
tdroxler: yes we could as when we start syncing we step back few hours to be sure we don't miss a block, so we do re-insert.

polarker:
Is it possible to update this PR based on Simer's batch write PR? So we will be sure that address table will benefit from batch write as well

tdroxler:
> Is it possible to update this PR based on Simer's batch write PR? So we will be sure that address table will benefit from batch write as well

Yes, I'll wait we merge the batch write PR and then merge it here and adapt the code

app/src/main/scala/org/alephium/explorer/persistence/queries/InputsQueries.scala
tdroxler: I improved [here](https://github.com/alephium/explorer-backend/pull/149/commits/d3fdfa7f7d38b19c0d70a23641276194c7a992e5) 

`Map.get` is O(1) right?

tdroxler:
I made the changes according to latest master, merged everything. It's ready for a final review.
I just launched a full sync with the mainnet to check everything is fine.

tdroxler:
@polarker this one needs some love :)

tdroxler:
I'll merge this one and we can come back on it if we need before deloying, it's a long PR and lot of further work depend on it

PR #153, opened 2022-02-25
Use `Long` db type for `TimeStamp`

tdroxler:
@polarker let's make the change of: "which interval to average" we discussed in an other PR, I will need to rework the hashrates' queries + fix all the `HashrateServiceSpec`.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
@@ -16,9 +16,10 @@
 
 package org.alephium.explorer.persistence.queries
 
-import java.time._
+import scala.concurrent.ExecutionContext
 
 import slick.basic.DatabaseConfig
+import slick.dbio.DBIOAction
 import slick.jdbc.JdbcProfile
 
 import org.alephium.explorer.persistence._
@@ -33,33 +34,42 @@ trait HashrateQueries extends CustomTypes {
   def getHashratesQuery(from: TimeStamp,
                         to: TimeStamp,
                         interval: Int): DBActionSR[(TimeStamp, BigDecimal)] = {
-    val fromInstant = Instant.ofEpochMilli(from.millis)
-    val toInstant   = Instant.ofEpochMilli(to.millis)
 
     sql"""
         SELECT timestamp, value
         FROM hashrates
-        WHERE interval_type = #$interval
-        AND timestamp >= '#${fromInstant.toString}'
-        AND timestamp <= '#${toInstant.toString}'
+        WHERE interval_type = $interval
+        AND timestamp >= $from
+        AND timestamp <= $to
         ORDER BY timestamp
       """.as[(TimeStamp, BigDecimal)]
   }
 
   def computeHashratesAndInsert(from: TimeStamp, intervalType: Int): DBActionW[Int] = {
-    val hashrates = (intervalType match {
-      case 0 => compute10MinutesHashrateRawString(from)
-      case 1 => computeHourlyHashrateRawString(from)
-      case 2 => computeDailyHashrateRawString(from)
-      case _ => "()"
-    })
+    if (intervalType >= 0 || intervalType <= 2) {
+      val dateGroup = (intervalType match {
+        case 0 => tenMinutesQuery
+        case 1 => hourlyQuery
+        case 2 => dailyQuery
+        case _ => ""
+      })
 
-    sqlu"""
-      INSERT INTO hashrates (timestamp, value, interval_type)
-      (#$hashrates)
-      ON CONFLICT (timestamp, interval_type) DO UPDATE
-      SET value = EXCLUDED.value
-    """
+      sqlu"""
+        INSERT INTO hashrates (timestamp, value, interval_type)
+        SELECT
+        EXTRACT(EPOCH FROM ((#$dateGroup) AT TIME ZONE 'UTC')) * 1000 as ts,
+        AVG(hashrate),
+        $intervalType
+        FROM block_headers
+        WHERE timestamp >= $from
+        AND main_chain = true
+        GROUP BY ts
+        ON CONFLICT (timestamp, interval_type) DO UPDATE
+        SET value = EXCLUDED.value
+      """
+    } else {
+      DBIOAction.successful(0)
+    }
   }
 
   /*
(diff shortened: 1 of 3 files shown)


app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -33,33 +34,42 @@ trait HashrateQueries extends CustomTypes {
...
-    })
-
-    sqlu"""
-      INSERT INTO hashrates (timestamp, value, interval_type)
-      (#$hashrates)
-      ON CONFLICT (timestamp, interval_type) DO UPDATE
-      SET value = EXCLUDED.value
-    """
+    if (intervalType >= 0 || intervalType <= 2) {
+      val dateGroup = (intervalType match {
```
polarker, line 50: Can we use proper type for `intervalType`? At least introduce constants with meaningful names

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -33,33 +34,42 @@ trait HashrateQueries extends CustomTypes {
...
+        case 0 => tenMinutesQuery
```
polarker, line 51: I am wondering if `10 minutes` average makes sense now

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -33,33 +34,42 @@ trait HashrateQueries extends CustomTypes {
...
+      })
+
+      sqlu"""
+        INSERT INTO hashrates (timestamp, value, interval_type)
+        SELECT
+        EXTRACT(EPOCH FROM ((#$dateGroup) AT TIME ZONE 'UTC')) * 1000 as ts,
+        AVG(hashrate),
+        $intervalType
+        FROM block_headers
+        WHERE timestamp >= $from
```
polarker, line 64: shall the condition be `timestamp >= $from && timestamp < ${from + interval}`?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -33,33 +34,42 @@ trait HashrateQueries extends CustomTypes {
...
 
     sql"""
         SELECT timestamp, value
         FROM hashrates
-        WHERE interval_type = #$interval
-        AND timestamp >= '#${fromInstant.toString}'
-        AND timestamp <= '#${toInstant.toString}'
+        WHERE interval_type = $interval
+        AND timestamp >= $from
+        AND timestamp <= $to
```
polarker, line 43: shall the condition be `[from, to)`, half close and half open interval?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -76,58 +86,70 @@ trait HashrateQueries extends CustomTypes {
...
+  private def computeHashrateRawString(from: TimeStamp, dateGroup: String, intervalType: Int) = {
+    sql"""
         SELECT
-        $dateGroup as date,
+        EXTRACT(EPOCH FROM ((#$dateGroup) AT TIME ZONE 'UTC')) * 1000 as ts,
         AVG(hashrate),
         $intervalType
         FROM block_headers
-        WHERE timestamp >= '${instant.toString}'
+        WHERE timestamp >= $from
```
polarker, line 150: same here

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -76,58 +86,70 @@ trait HashrateQueries extends CustomTypes {
    * round to above 10 minutes: ceiling(26.09 / 10) * 10 = 30
    * result: 8:30
    */
-  private def compute10MinutesHashrateRawString(from: TimeStamp): String = {
+
+  private val timestampTZ = "to_timestamp(timestamp/1000.0) AT TIME ZONE 'UTC'"
```
polarker, line 90: `to_timestamp` converts seconds to date?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -33,33 +34,42 @@ trait HashrateQueries extends CustomTypes {
...
+    if (intervalType >= 0 || intervalType <= 2) {
+      val dateGroup = (intervalType match {
+        case 0 => tenMinutesQuery
+        case 1 => hourlyQuery
+        case 2 => dailyQuery
+        case _ => ""
+      })
+
+      sqlu"""
+        INSERT INTO hashrates (timestamp, value, interval_type)
```
polarker, line 58: Shall we store the hashrates as one table or several separate tables?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
```diff
@@ -16,9 +16,10 @@
 
 package org.alephium.explorer.persistence.queries
 
-import java.time._
+import scala.concurrent.ExecutionContext
```
polarker, line 19: Why `ExecutionContext` was not needed before?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: yes it's on my TODO list, but thought it was not linked to that PR

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: As we have an index on the interval_type, I'm not sure it's worth it.

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: no, the `from` here is to avoid re-computing  hashrates every time from the first block. Before running the request we check when was the latest hashrates computed, we step back and start from that.

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: yes, but seconds can have a float value to get the ms, that's why I divide by `1000.0` and not `1000`:
```
postgres=# select to_timestamp( 1646056461886/1000);
      to_timestamp      
------------------------
 2022-02-28 13:54:21+00
(1 row)

postgres=# select to_timestamp( 1646056461886/1000.0);
        to_timestamp        
----------------------------
 2022-02-28 13:54:21.886+00
(1 row)
```

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: to do the `map` here https://github.com/alephium/explorer-backend/blob/7cbf417d4e9408536fbb3a0ece59feb905b82208/app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala#L113 but I realize those functions using that need the EC are just for testing purpose, I'll clean that, nice catch

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
polarker: Is it possible that this query will return a huge number of elements?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
polarker: Since second is precise enough for all of your queries, `timestamp / 1000` is much more efficient

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
polarker: Since we created index for `interval_type`, it is fine

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: easy to compute, if you need to compute from scratch the 10 minutes interval and let say Alephium run since 1 year:
`6 * 24 * 365 = 52560`
but the "from scratch" case is computed only once, as we then to what I described just above.

polarker: sounds good to me!

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: mmm again it's a matter of point of view :)  if I request hashrates from 08:00 to 14:00 I would expect to have both 08:00 and 14:00 in the data I receive.

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: let's tackle this with https://github.com/alephium/explorer-backend/issues/154 in another PR as it breaks most of the tests

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: let's keep it like this for now, this will be anyway fixed with https://github.com/alephium/explorer-backend/issues/154

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
polarker: @tdroxler Let's remove 10 minutes for now. We both think it's not helpful

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
polarker: okay, then let's use close interval

polarker: the main issue is that if one queries for several sequential interval, some hashrate would appear in two queries

app/src/main/scala/org/alephium/explorer/persistence/schema/HashrateSchema.scala
```diff
@@ -36,13 +35,13 @@ trait HashrateSchema extends CustomTypes {
 
     def pk: PrimaryKey = primaryKey("hashrates_pk", (timestamp, intervalType))
 
+    def timestampIdx: Index    = index("hashrates_timestamp_idx", timestamp)
+    def intervalTypeIdx: Index = index("hashrates_interval_type_idx", intervalType)
```
polarker, line 40: We need to create partial index for timestamp I believe, i.e. `CREATE INDEX hashrates_timestamp_idx on timestamp where intervalType = 0`.

@simerplaha Do you think my comment makes sense? I remember you have built partial indexes for other tables already

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: ah good point,  ok so let's say if you request  08:00 to 14:00, you'll get 8:00 and 13:00

app/src/main/scala/org/alephium/explorer/persistence/schema/HashrateSchema.scala
simerplaha: Yep that makes sense. Just tried looking if I had benchmarks documented anywhere for adding partial index (`main_chain = true`) to `block_headers` table ([PR:131](https://github.com/alephium/explorer-backend/pull/131)) but I think I lost it. My bad :(. Couldn't find the code for the benchmark either. We only benchmark with or without the index. 

Benchmarking `hashrates_timestamp_idx ` index would tell us the truth. Please let me know if you'd like me to run the benchmark?

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: let's keep it like this for now and we'll see if someone complains

app/src/main/scala/org/alephium/explorer/persistence/schema/HashrateSchema.scala
tdroxler: if you have the courage, yes please that would be super kind :heart:

tdroxler: we will actually do it in another task: https://github.com/alephium/explorer-backend/issues/158

app/src/main/scala/org/alephium/explorer/persistence/queries/HashrateQueries.scala
tdroxler: done

polarker (approved):
LGTM

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomTypes.scala
```diff
@@ -100,9 +100,8 @@ trait CustomTypes extends JdbcProfile {
   implicit lazy val intervalTypeType: JdbcType[IntervalType] =
     MappedJdbcType.base[IntervalType, Int](
       _.value, {
-        case 0 => IntervalType.TenMinutes
-        case 1 => IntervalType.Daily
-        case 2 => IntervalType.Hourly
+        case 0 => IntervalType.Daily
```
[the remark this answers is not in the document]
tdroxler: good idea, it's done and I did the same idea with the string representation

PR #165, opened 2022-03-13
Add doc check in CI

polarker:

PR #166, opened 2022-03-14
Generate scaladoc 

simerplaha:
## Overview

- PR for [comment](https://github.com/alephium/explorer-backend/pull/165#issuecomment-1066064429)
- Adds plugin `sbt-unidoc` to support multiple projects references in docs.
- References to Slick types link to https://www.javadoc.io/doc/com.typesafe.slick/slick_2.13/3.3.2 
- References to Scala types link to https://scala-lang.org/api/2.13.3

## Generating docs
`sbt unidoc`

See [readme.md](https://github.com/alephium/explorer-backend/blob/c5cb628d684c30eb3d1efd8968f2f3b3bed96ea3/README.md#scaladoc)


build.sbt
```diff
@@ -139,6 +206,16 @@ lazy val app = mainProject("app")
       case "module-info.class" =>
         MergeStrategy.discard
       case other => (assembly / assemblyMergeStrategy).value(other)
+    },
+    apiMappings ++= {
+      val slickAPIMapping =
```
tdroxler, line 211: we could maybe have this `val slickAPIMapping` define outside so we can reuse it here and below

tdroxler (approved):
Great :rocket: LGTM

polarker (approved):
Thanks ❤️

build.sbt
simerplaha: Yep that would've been much cleaner. But sbt doesn't allow it. 

> error: `value` can only be called on a task within a task definition macro
classPath    = (Compile / fullClasspath).value

I'm sure there is a way to make this work, it's Scala after all. Will try to figure it out another PR.

PR #168, opened 2022-03-15
Update common

tdroxler:
To be merged after https://github.com/alephium/dev-alephium/pull/623 + we need to update the `Dependencies.scala`

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
@@ -184,14 +184,14 @@ object BlockFlowClient {
         case (tx, txIndex) =>
           tx.contractInputs.toSeq.zipWithIndex.map {
             case (outputRef, index) =>
-              val shiftIndex = index + inputs.size
-              outputRefToInputToEntity(outputRef,
-                                       hash,
-                                       tx.unsigned.hash,
-                                       block.timestamp,
-                                       mainChain,
-                                       shiftIndex,
-                                       txIndex)
+              val shiftIndex = index + tx.unsigned.inputs.length
+              outputRefToInputEntity(outputRef,
+                                     hash,
+                                     tx.unsigned.hash,
+                                     block.timestamp,
+                                     mainChain,
+                                     shiftIndex,
+                                     txIndex)
           }
       }
     inputs ++ contractInputs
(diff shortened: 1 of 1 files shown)


app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -167,13 +167,80 @@ object BlockFlowClient {
...
+            case (in, index) =>
+              inputToEntity(in, hash, tx.unsigned.hash, block.timestamp, mainChain, index, txIndex)
+          }
+      }
+    val contractInputs =
+      transactions.flatMap {
+        case (tx, txIndex) =>
+          tx.contractInputs.toSeq.zipWithIndex.map {
+            case (outputRef, index) =>
+              val shiftIndex = index + inputs.size
```
polarker, line 187: should be `index + tx.unsigned.inputs.size` instead of `inputs.size`?

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -167,13 +167,80 @@ object BlockFlowClient {
...
+                             mainChain,
+                             txIndex)
+          }
+      }
+    val generatedOutputs =
+      transactions.flatMap {
+        case (tx, txIndex) =>
+          tx.generatedOutputs.toSeq.zipWithIndex.map {
+            case (out, index) =>
+              val shiftIndex = index + outputs.size
```
polarker, line 223: same here, we should use `index + tx.unsigned.outputs.size`

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -276,13 +331,33 @@ object BlockFlowClient {
                 txIndex)
   }
 
+  private def outputRefToInputToEntity(outputRef: api.model.OutputRef,
```
polarker, line 334: `outputRefToInputEntity`?

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
tdroxler: oooh nice catch, I used `inputs` thinking it's the same, but inputs is actually `transactions.flatMap(_.unsigned.inputs.size)`

PR #170, opened 2022-03-16
Set query parameter limit in `QuerySplitter` to 2000

simerplaha:
- PR for [comment](https://github.com/alephium/explorer-backend/pull/162#discussion_r826874299)
- Added a [test](https://github.com/alephium/explorer-backend/compare/set-query-limit-to-2000?expand=1#diff-cc6b9d839491e75f36acb610167b553a9e42cf7645bbd735e662bdc5d08c3593R159) following the comment.
- Renamed `QueryUtil` to `QuerySplitter`


app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala
```diff
@@ -23,22 +23,25 @@ import slick.dbio.DBIOAction
...
+  /** Maximum number of parameters per query.
+    *
+    * @note This value cannot be > [[scala.Short.MaxValue]].
+    *       See issue <a href="https://github.com/alephium/explorer-backend/issues/160">#160</a>.
     *
     * TODO: This should be passed as a configuration instead of hardcoding it
     *       and should be passed as parameter to the function using this value.
     * */
-  val postgresDefaultParameterLimit: Short = Short.MaxValue
+  val maxParametersPerQuery: Short = 2000
```
polarker, line 36: We could introduce `val maxRowPerQuery: Short = 200` to make the cache bounded by the number of rows, instead of number of parameters. `maxParametersPerQuery` would be `maxRowPerQuery * rowParams`

simerplaha: - `val maxRowPerQuery: Short = 200` added [here](https://github.com/alephium/explorer-backend/pull/170/files#diff-c5205d5c5aadb4be1f66145f2ff3256aa1aa61730abd1acbd2e2af75c2bb201dR39)
- `maxRowPerQuery * rowParams` added [here](https://github.com/alephium/explorer-backend/pull/170/files#diff-c5205d5c5aadb4be1f66145f2ff3256aa1aa61730abd1acbd2e2af75c2bb201dR78)

app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala
```diff
@@ -0,0 +1,142 @@
...
+    *                              - _3 = Previous query. Used to join with next query.
+    * @tparam R type of rows
+    * @tparam Q type of query
+    * @return A single batched query
+    */
+  def splitFoldLeftLimitByMaxParams[R, Q](
+      rows: Iterable[R],
+      initialQuery: Q,
+      columnsPerRow: Short,
+      maxParamsPerQuery: Short)(foldLeft: (Iterable[R], String, Q) => Q): Q = {
```
polarker, line 114: If we pass by `maxRowsPerQuery` instead of `maxParamsPerQuery`, then we will not recalculate `maxRows` for this function.

We could use `val maxRowsPerQueryAllowed =
      maxRowsPerQuery min postgresDefaultParameterLimit.toInt / columnsPerRow`

polarker (approved):
Definitely LGTM ❤️ Feel free to ignore my nitpicking comment

PR #171, opened 2022-03-17
Feature/better open api

tdroxler:


polarker:
Since I updated full node OpenApi a lot, this needs to be updated too

app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
```diff
@@ -18,7 +18,13 @@ package org.alephium.explorer.service
...
 import scala.concurrent.{ExecutionContext, Future}
 
-import org.alephium.explorer.api.model.{Address, Pagination, Transaction, TransactionLike}
+import org.alephium.explorer.api.model.{
+  Address,
+  ConfirmedTransaction,
+  Pagination,
+  Transaction,
+  TransactionLike
+}
```
polarker, line 27: Nit: we could collapse the imports

app/src/test/scala/org/alephium/explorer/api/model/ApiModelSpec.scala
```diff
@@ -48,6 +47,23 @@ class ApiModelSpec() extends AlephiumSpec with Generators {
     }
   }
 
+  it should "ConfirmedTransaction" in {
+    forAll(transactionGen) { tx =>
+      val expected = s"""
+       |{
+       |  "type": "confirmed",
```
polarker, line 54: Let's use capitalized word for types, which is the case for most languages. In JavaScript, classes are named in `Foo` as far as I see

polarker (approved):
LGTM

PR #178, opened 2022-03-28
Add `/infos/total-transactions` and `/infos/average-block-times` endpoints

tdroxler:
cc @lf94

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
@@ -36,7 +36,7 @@ import org.alephium.explorer.persistence.queries.InputQueries._
 import org.alephium.explorer.persistence.schema._
 import org.alephium.protocol.config.GroupConfig
 import org.alephium.protocol.model.ChainIndex
-import org.alephium.util.TimeStamp
+import org.alephium.util.{Duration, TimeStamp}
 
 trait BlockDao {
   def get(hash: BlockEntry.Hash): Future[Option[BlockEntry]]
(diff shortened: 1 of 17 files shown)


app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
```diff
@@ -95,7 +95,40 @@ object BlockDao {
...
+      } else {
+        Duration.zero
+      }
+    }
+
+    private val cachedBlockTimes: AsyncLoadingCache[ChainIndex, Duration] = Caffeine
+      .newBuilder()
+      .maximumSize(groupConfig.chainNum.toLong)
+      .expireAfterWrite(5, java.util.concurrent.TimeUnit.SECONDS)
+      .buildAsync(blockTimeAsyncLoader)
```
simerplaha, line 131: Since `cachedBlockTimes` and `latestBlockAsyncLoader` have the same configuration and keys do you think we could use a single cache or do they have to be seperate?

I'm totally ok with this. I couldn't tell single which way is better without profiling :|

simerplaha (approved):
Awesome!!!! 🚀 🚀

app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
simerplaha, line 131: Since `cachedBlockTimes` and `latestBlockAsyncLoader` have the same configuration and key type do you think we could use a single cache or do they have to be seperate?

Seperate looks fine as well. I couldn't tell which way is better without profiling :|

simerplaha (commented):
Fixed comment

app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
tdroxler: ah haven't thought about this, but for me it really serves two different purposes so it's clearer in my mind to have two different caches, the fact that they have the same key and config is more like a coincidence + they have very different functions to update their cache, I would be afraid to slow down one of them by merging, but as you said without profiling it's hard to tell.

PR #179, opened 2022-03-28
Rename order/index table fields

### LINKED ISSUE #157: Rename `order` table fields
`order` is a sql key-word, we need to make sore we don't use it. In few place we also use `index`...which is also a key word ^^ let's use something meaningful: `tx_order` etc

tdroxler:
Thus are reserved SQL key words and we had some issues in the past.

Resolves #157

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/model/TransactionEntity.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/model/TransactionEntity.scala
@@ -27,6 +27,6 @@ final case class TransactionEntity(
     chainTo: GroupIndex,
     gasAmount: Int,
     gasPrice: U256,
-    index: Int,
+    order: Int,
     mainChain: Boolean
 )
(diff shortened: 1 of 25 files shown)


simerplaha (approved):
🚀 🚀

app/src/main/scala/org/alephium/explorer/persistence/model/TransactionEntity.scala
```diff
@@ -27,6 +27,6 @@ final case class TransactionEntity(
     chainTo: GroupIndex,
     gasAmount: Int,
     gasPrice: U256,
-    index: Int,
+    order: Int,
```
polarker, line 30: Should we avoid the order here as well?

tdroxler: no in the scala entity it's all fine and cleaner, rather than repeating `Transaction.tx_order`, `Input.input_order` etc

Issue #156, opened 2022-02-28
Removal of typed query `listMainChain`


simerplaha:
Task for [PR comment](https://github.com/alephium/explorer-backend/pull/131#discussion_r795686087).

PR #131 replaces `listMainChain` with `listMainChainSQL` without removing `listMainChain` so it could be used for [benchmarking](https://github.com/alephium/explorer-backend/blob/master/benchmark/src/main/scala/org/alephium/explorer/benchmark/db/state/ListBlocksReadState.scala) and [a test](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/persistence/dao/BlockDaoSpec.scala#L92). 

Since the PR is merged. Code related to `listMainChain`, benchmarks and that one test can be removed.

simerplaha:
PR #254 from Thomas resolved this.

PR #180, opened 2022-03-29
Cached `main_chain` count query

### LINKED ISSUE #119: `select count(*) ...` queries on partial indexes should be constant
# Overview

Count queries like the following on indexes that fully satisfy the where clause should be constant time.

### Query
```sql
select count(*) from block_headers where main_chain = true;
```

### Index
```sql
create index blocks_main_chain_idx on block_headers (main_chain) where main_chain = true;
```

# Issue

The query is not constant and is doing aggregate scan.

simerplaha:
- Added a cache `cacheRowCount: CaffeineAsyncCache[Query[_, _, Seq], Int]` which will cache any `count` query. This PR uses it for `listMainChain` only.
- Added a wrapper `CaffeineAsyncCache` to centralise handling Caffeine's Java to Scala type conversions. 
- Resolves #119

# Benchmark

The following benchmark shows around 20% performance improvement. 

image

# Cleanup

#156 will remove deprecated apis `listMainChain` and `listMainChainSQL` with `listMainChainSQLCached`.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
@@ -16,6 +16,8 @@
 
 package org.alephium.explorer.persistence.dao
 
+import java.util.concurrent.{CompletableFuture, Executor}
+
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 import scala.jdk.FutureConverters._
@@ -29,6 +31,7 @@ import slick.jdbc.PostgresProfile.api._
 
 import org.alephium.explorer.AnyOps
 import org.alephium.explorer.api.model._
+import org.alephium.explorer.cache.CaffeineAsyncCache
 import org.alephium.explorer.persistence._
 import org.alephium.explorer.persistence.model._
 import org.alephium.explorer.persistence.queries._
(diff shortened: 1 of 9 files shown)


app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
```diff
@@ -130,6 +133,47 @@ object BlockDao {
...
+    private val cacheRowCount: CaffeineAsyncCache[Query[_, _, Seq], Int] =
+      CaffeineAsyncCache {
+        Caffeine
+          .newBuilder()
+          .refreshAfterWrite(2, java.util.concurrent.TimeUnit.MINUTES)
+          .buildAsync[Query[_, _, Seq], Int] {
+            new AsyncCacheLoader[Query[_, _, Seq], Int] {
+              override def asyncLoad(table: Query[_, _, Seq],
+                                     executor: Executor): CompletableFuture[Int] =
+                run(table.length.result).asJava.toCompletableFuture
```
tdroxler, line 167: I was first surprise to see the query to be the key of the cache, but in a sense it makes more sense that what I did to have a "fake" key : https://github.com/alephium/explorer-backend/blob/48986dd9b346aa26a05eb08f3016e14858a8e742/app/src/main/scala/org/alephium/explorer/persistence/dao/TransactionDao.scala#L55-L58 Can you maybe abstract what you did there so we could reuse it in the `TransactionDao` too?

app/src/test/scala/org/alephium/explorer/persistence/dao/BlockDaoSpec.scala
```diff
@@ -169,10 +171,108 @@ class BlockDaoSpec extends AlephiumSpec with ScalaFutures with Generators with E
...
+        blockDao
+          .listMainChainSQLCached(Pagination.unsafe(0, 1, Random.nextBoolean()))
+          .futureValue
+          ._2 is expectedMainChainCountTotal
+        //Dispatch a query so the cache get populated
+        blockDao
+          .getRowCountFromCacheIfPresent(blockDao.mainChainQuery)
+          .map(_.futureValue) is Some(expectedMainChainCountTotal)
+    }
+  }
```
tdroxler, line 270: super complet testing :+1:

tdroxler (approved):
LGTM, regarding the abstration of the row count cache that we could reuse in `TxDao`, feel free to create a ticket or to tackle in this PR

simerplaha:
> LGTM, regarding the abstration of the row count cache that we could reuse in TxDao, feel free to create a ticket or to tackle in this PR

Yep. Sounds good. Created ticket #181 for another PR :)

app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala
simerplaha: Yep :) [issue](https://github.com/alephium/explorer-backend/pull/180#issuecomment-1082603352) created

PR #182, opened 2022-03-30
#111: enable inlining

### LINKED ISSUE #111: Inline `GroupIndex` and `Height`
These types are wrappers around primitive type (`Int`) and can be inlined.

simerplaha:
- Enables inlining
- Resolves #111


simerplaha:
Discussed this with Thomas.

Will reopen this PR again after confirming the advantage of inlining over value classes.

polarker:
The `inliningOptions` looks interesting ❤️  I will play with it in the future for full node

PR #185, opened 2022-03-31
Add `/address/total-transaction` and `/address/balance` endpoint

tdroxler:
The idea is for the front-end to call those two endpoints rather than
the one to get all info at once.

With this we could show the address details faster and also the
transactions pagination.


simerplaha (approved):
🚀 🚀

Issue #186, opened 2022-03-31
Block `main_chain` count query should populate in read-only


simerplaha:
PR #180 (Cached `main_chain` count query) does not solve for `read-only` mode.

tdroxler:
One idea could be to add those timeout value in the config file and pass different value if it's read-only or not.

polarker:
Why the cache is different for `read-only` and `write` mode?

tdroxler:
for the block count, in the `write` mode, when we insert new blocks we invalidate the cache. We thus had a long timeout, because we knew the cache would be automatically re-newed on insert. But it doesn't work for the `read-only` case, as we don't insert, we don't invalidate.

simerplaha:
Hey @tdroxler, looking at this now. Think I've forgotten something about this issue from our chat. Can I please get a reminder of the problem here again?

Ran a test locally and here is what I understand so far.

In `readOnly` mode the `syncService` is disabled which means no writes are occurring right? Since the state of the database is not changing there shouldn't be a need to invalidate the cache?

Local test: I get 9413 for `select count(*) from block_headers where main_chain;` which is same in API.

![image](https://user-images.githubusercontent.com/1773953/162370006-82507758-6528-4731-85a2-6f7fe3a523b4.png)

Maybe in `readOnly` mode automatic expiration can be disabled all together?

simerplaha:
Ok, I'm guessing there are two services running on the same database, one is `readOnly` and other is `readAndWrite`. In that case the cache in `readOnly` mode instance would need invalidating the count?

tdroxler:
Yes both app are running on the same database. The `readOnly` get its cache invalidate by a timeout, so it's fine. I think I remember for the `read-write` app we were also invalidating while inserting new blocks right?
We discuss this with Cheng and we think we could afford to also let the `read-write` get invalidated by a timeout, no need to manually invalidate on insert. also we could raise the 5 seconds timeout, we don't need to have that level of live update. The questions is: what's a good timeout value? let's maybe start with 1 minutes and see how it performs.
The mean idea is to keep the same behavior for both read-only and read-write (just the sync services difference).
 WDYT?

simerplaha:
> we were also invalidating while inserting new blocks right?

Yep. After new blocks are inserted.

> We discuss this with Cheng and we think we could afford to also let the read-write get invalidated by a timeout, no need to manually invalidate on insert.

[...]

PR #195, opened 2022-04-06
Change block count cache timeout

tdroxler:
2 minutes is too long for the read-only app.

This is a quick fix related to #186 
An option we can check for later is to have all those timeouts set in
the config file and we can give different value between the read-only
and read-write app.

### DIFF
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -660,6 +660,198 @@
         }
       }
     },
+    "/addresses/{address}/total-transactions": {
+      "get": {
+        "tags": [
+          "Addresses"
+        ],
+        "description": "Get total transactions of a given address",
+        "operationId": "getAddressesAddressTotal-transactions",
+        "parameters": [
+          {
+            "name": "address",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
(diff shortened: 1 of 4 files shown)


nop33 (approved):

app/src/main/resources/explorer-backend-openapi.json
```diff
@@ -1429,7 +1429,7 @@
         "operationId": "putUtilsSanity-check",
         "responses": {
           "200": {
-
+
```
polarker, line 1624: I did not know that it generates an empty line here 😅

simerplaha (approved):
Nice! 🚀 

Just something to think about (probably should've mentioned in #119): We might be polling excessively. Now we will have 4 caches polling the database in short intervals of 5.seconds (possibly more caches to be added in the future). So on average, nearly every second we are dispatching a DB query which themselves are expensive and take time to respond. The sync process is also polling every 5.seconds. The UI is also polling around 5.seconds, a decent amount of idle browser sessions and we might not do so well in terms of performance unless we get bigger machines. We are polling everywhere!

tdroxler:
I fully aggree we need to be careful about polling, but for the caches we are not polling the db every 5 seconds, we are invalidating the cache every 5 seconds.

But you are right we need to think more about this, especially with the read-only app.

PR #197, opened 2022-04-13
#146: close test database connections after use

### LINKED ISSUE #146: Database connection created by tests should be closed after each run
# Issue 
Our test cases do not close connections created after each test run. 

We have a total of 70 test cases so it not a problem yet.

# Test
Quick test demo the error 

```scala
(1 to 300) foreach {
  _ =>
    val fixture = new Fixture {}
    import fixture.config.profile.api._
    fixture.run(fixture.blockDepsTable.result).futureValue
}
```
Errors the following after 214 iterations.

> java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached
[49.957s][warning][os,thread] Failed to start thread - pthread_create failed (EAGAIN) for attributes: stacksize: 1024k, guardsize: 4k, detached.

simerplaha:
Resolves #146

### DIFF
--- a/app/src/test/scala/org/alephium/explorer/ApplicationSpec.scala
+++ b/app/src/test/scala/org/alephium/explorer/ApplicationSpec.scala
@@ -38,7 +38,7 @@ import org.alephium.api.{ApiError, ApiModelCodec}
 import org.alephium.api.model
 import org.alephium.explorer.AlephiumSpec
 import org.alephium.explorer.api.model._
-import org.alephium.explorer.persistence.DatabaseFixture
+import org.alephium.explorer.persistence.DatabaseFixtureForAll
 import org.alephium.explorer.persistence.model.BlockEntity
 import org.alephium.explorer.service.BlockFlowClient
 import org.alephium.explorer.util.TestUtils._
@@ -51,7 +51,7 @@ trait ApplicationSpec
     extends AlephiumSpec
     with ScalatestRouteTest
     with ScalaFutures
-    with DatabaseFixture
+    with DatabaseFixtureForAll
     with Generators
     with Eventually
     with UpickleCustomizationSupport {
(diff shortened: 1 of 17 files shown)


app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
```diff
@@ -23,7 +23,7 @@ import slick.jdbc.PostgresProfile
 import slick.jdbc.PostgresProfile.api._
 
 trait DBRunner {
-  val databaseConfig: DatabaseConfig[PostgresProfile]
+  def databaseConfig: DatabaseConfig[PostgresProfile]
```
tdroxler, line 26: shouldn't we have a `lazy val` instead? in case someone is calling twice `databaseConfig`

tdroxler (approved):
Niiiice! LGTM, one minor comment

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
simerplaha: Please correct me if I'm wrong but I don't think traits can have `lazy` definitions. 

If it does then the object or class that implements this trait would define it as `val`, `var`, `lazy val` or `def` and its definition would get used during runtime not the one in the trait.

simerplaha: Ah I think you know this stuff already. In summary it shouldn't need a lazy val 🤔

polarker (approved):
LGTM ❤️

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
tdroxler: ah stupid me, I didn't notice it was only the definition in the trait :face_palm:

PR #199, opened 2022-04-19
Feature/finalizer

tdroxler:
maim-1650372676
:rocket: :rocket: :rocket: 

The update of the outputs is quite heavy, especially the first time. Running one query is just too long, you don't know how long it will take and if the process fails in the middle for any reason you need do restart the all query. 
One recommended way I saw in multiple place was to store the result you want to update in a temporary table and then update by batch reading this table. I took inspiration from [this article](https://fullstackcode.dev/2021/10/02/updating-millions-of-records-in-batches-in-postgresql/).

Using temporary table works great for a one shot update, but as we are updating regularly, my database session didn't like it and the explorer-backend was crashing we trying to use the temporary table a second time, so I now create a real table and just truncate it when done with the update.

Also using a procedure works great inside psql, but I couldn't get the logging to come up the console, so I went the scala way for the loop of batch update, otherwise you might think the app is stuck when running the procedure.

This PR also introduce a very minimalist migration system, let's try to improve it in the future, this will save us a lot of syncing time.

I'm not fully happy of this PR, but let's start to review/discuss it as it's a big piece and it start to be urgent.


simerplaha:
Thats fast! 🚀 🚀

app/src/main/scala/org/alephium/explorer/persistence/model/AppState.scala
```diff
@@ -0,0 +1,51 @@
...
+    override def _deserialize(input: ByteString): SerdeResult[Staging[Value]] =
+      byteSerde._deserialize(input).flatMap {
+        case Staging(byte, rest) =>
+          if (byte == 0) {
+            serdeTS.deserialize(rest).map(ts => Staging(Time(ts), ByteString.empty))
+          } else {
+            Left(SerdeError.wrongFormat(s"Invalid AppState.Value"))
+          }
+      }
+  }
```
simerplaha, line 50: Very minor update so we don't have to type-cast in the `CustomGetResult`.

```scala
sealed trait Value
... 

object Time extends Serde[AppState.Time] {

  override def serialize(input: Time): ByteString =
    ByteString(0) ++ serde.serialize(ts.value)

  override def _deserialize(input: ByteString): SerdeResult[Staging[Time]] =
    byteSerde._deserialize(input).flatMap {
      case Staging(byte, rest) =>
        if (byte == 0) {
          serdeTS.deserialize(rest).map(ts => Staging(Time(ts), ByteString.empty))
        } else {
          Left(SerdeError.wrongFormat(s"Invalid AppState.Time"))
        }
    }
}

object Value extends Serde[AppState.Value] {

  override def serialize(input: Value): ByteString =
    input match {
      case time: Time => Time.serialize(time)
    }

  override def _deserialize(input: ByteString): SerdeResult[Staging[Value]] =
    Time.deserialize(input) map (Staging(_, ByteString.empty))

}
```

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomGetResult.scala
```diff
@@ -80,6 +80,21 @@ object CustomGetResult {
       U256.unsafe(result.nextBigDecimal().toBigInt.bigInteger)
     }
 
+  @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
+  implicit lazy val appStateValueGetResult: GetResult[AppState.Value] =
+    (result: PositionedResult) =>
+      AppState.Value.deserialize(ByteString.fromArrayUnsafe(result.nextBytes())).toOption.get
```
simerplaha, line 86: Should we replace `.toOption.get` with throwing the `SerdeError` to get actual stack-trace in the logs in case of failure?

```scala
AppState.Value
  .deserialize(ByteString.fromArrayUnsafe(result.nextBytes()))
  .fold(throw _, value => value)
```

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomGetResult.scala
```diff
@@ -80,6 +80,21 @@ object CustomGetResult {
...
+
+  @SuppressWarnings(
+    Array("org.wartremover.warts.OptionPartial", "org.wartremover.warts.AsInstanceOf"))
+  implicit lazy val appStateTimeGetResult: GetResult[AppState.Time] =
+    (result: PositionedResult) =>
+      AppState.Value
+        .deserialize(ByteString.fromArrayUnsafe(result.nextBytes()))
+        .toOption
+        .get
+        .asInstanceOf[AppState.Time]
```
simerplaha, line 96: Same as above for `toOption.get`. 

Following the minor change in the Serialiser code above to remove this typecast `.asInstanceOf[AppState.Time]`.

```scala
implicit val appStateTimeGetResult: GetResult[AppState.Time] =
  (result: PositionedResult) =>
    AppState.Time
      .deserialize(ByteString.fromArrayUnsafe(result.nextBytes()))
      .fold(throw _, value => value)
```

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomJdbcTypes.scala
```diff
@@ -101,4 +102,10 @@ object CustomJdbcTypes {
       IntervalType.unsafe
     )
 
+  @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
+  implicit lazy val appStateValue: JdbcType[AppState.Value] =
+    MappedJdbcType.base[AppState.Value, ByteString](
+      appState => AppState.Value.serialize(appState.value),
+      input    => AppState.Value.deserialize(input).toOption.get
+    )
```
simerplaha, line 110: Similar to above, instead of `.toOption.get` should we use `fold` or `match` instead to get the actual error in the call stack in case of failure?

app/src/main/scala/org/alephium/explorer/persistence/queries/BlockDepQueries.scala
```diff
@@ -40,28 +40,29 @@ object BlockDepQueries {
       DBIOAction.successful(0)
     } else {
       //generate '?' placeholders for the parameterised SQL query
-      val placeholder = paramPlaceholder(rows = deps.size, columns = 3)
-      val query =
-        s"""
+      QuerySplitter.splitUpdates(rows = deps, columnsPerRow = 3) { (deps, placeholder) =>
```
simerplaha, line 43: `QuerySplitter.splitUpdates` already checks for empty inputs [here](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala#L123). The [test-case](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/persistence/queries/QuerySplitterSpec.scala#L66). So the check above for `if(deps.isEmpty)` is probably not needed?

app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
```diff
@@ -217,10 +217,9 @@ trait BlockQueries extends TransactionQueries with StrictLogging {
     if (blocks.isEmpty) {
       DBIOAction.successful(0)
```
simerplaha, line 218: Similar to above. With `QuerySplitter.splitUpdates` this check is no longer needed?

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -56,10 +56,10 @@ trait TransactionQueries extends StrictLogging {
     if (transactions.isEmpty) {
       DBIOAction.successful(0)
```
simerplaha, line 57: Similar to above. With `QuerySplitter.splitUpdates` this check is no longer needed?

simerplaha (changes requested):
Awesome PR! 🚀 

Just a question: We've used `lazy val` at a few places here. I'm wondering if we need them at all since there is no concurrent mutation or expensive computation involved. `lazy val` generates code wrapped in `synchronized` blocks with `volatile` variable right (last time I checked for 2.12, 2.13 might be different)? So there would be a runtime cost for accessing these vals.

app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -103,14 +106,34 @@ class Application(
...
+
+  private def stopReadWrite(): Future[Unit] = {
+    if (!readOnly) {
+      blockFlowSyncService.stop()
+    } else {
+      Future.successful(())
+    }
+  }
+
   def start: Future[Unit] = {
```
polarker, line 133: Future Improvements: We should either use or build a simple framework for services so they can be started and stopped properly without having to deal with dependencies manually. Full node ref: https://github.com/alephium/alephium/blob/master/util/src/main/scala/org/alephium/util/Service.scala

app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -103,14 +106,34 @@ class Application(
...
     for {
       _       <- healthCheckDao.healthCheck()
-      _       <- dbInitializer.createTables()
-      _       <- if (readOnly) Future.successful(()) else startSyncService()
+      _       <- startReadWrite()
```
polarker, line 136: I am not sure about the naming `startReadWrite`?

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
```diff
@@ -51,10 +52,21 @@ class DBInitializer(val databaseConfig: DatabaseConfig[PostgresProfile])(
...
-      TransactionPerAddressSchema.table
+      TransactionPerAddressSchema.table,
+      AppStateSchema.table
     )
 
-  def createTables(): Future[Unit] = {
+  def initialize(): Future[Unit] = {
+    for {
+      _ <- createTables()
+      _ <- Migrations.migrate(databaseConfig)
```
polarker, line 62: Does this mean the migration will run every time we start the application?

We will probably introduce a version to the system and run migration based on the version.

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
```diff
@@ -51,10 +52,21 @@ class DBInitializer(val databaseConfig: DatabaseConfig[PostgresProfile])(
...
+      _ <- createIndexes()
```
polarker, line 63: Do we want to put index creation before the migration?

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
```diff
@@ -32,3 +32,13 @@ trait DBRunner {
       case error => throw new RuntimeException(error)
     }
 }
+
+object DBRunner {
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  def run[R, E <: Effect](databaseConfig: DatabaseConfig[PostgresProfile])(action: DBAction[R, E])(
+      implicit executionContext: ExecutionContext): Future[R] =
+    databaseConfig.db.run(action).recover {
+      case error => throw new RuntimeException(error)
```
polarker, line 41: Is it common practice to throw RuntimeException for any DB error? cc @simerplaha

app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import com.typesafe.scalalogging.StrictLogging
+import slick.basic.DatabaseConfig
+import slick.jdbc.PostgresProfile
+import slick.jdbc.PostgresProfile.api._
+
+import org.alephium.explorer.persistence._
+
+object Migrations extends StrictLogging {
+  val outputStatus: DBActionW[Int] = sqlu"""
```
polarker, line 29: How about naming it as `addOutputStatusColumn`?

app/src/main/scala/org/alephium/explorer/persistence/model/AppState.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.alephium.serde._
+import org.alephium.util.TimeStamp
+
+final case class AppState(
+    key: String,
+    value: AppState.Value
+)
+object AppState {
+
+  trait Value
```
polarker, line 31: Nit: I personally think this abstraction is over-engineering, as the values of AppState are not correlated. Therefore a pure ByteString might be better. We then let each state to handle encoding/decoding separately. Here, we should really provide a key-value table.

I am fine with this solution, but it's again KISS principle.

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
```diff
@@ -40,6 +41,8 @@ object OutputSchema extends SchemaMainChain[OutputEntity]("outputs") {
     def lockTime: Rep[Option[TimeStamp]] = column[Option[TimeStamp]]("lock_time")
     def outputOrder: Rep[Int]            = column[Int]("output_order")
     def txOrder: Rep[Int]                = column[Int]("tx_order")
+    def spentFinalized: Rep[Option[Transaction.Hash]] =
+      column[Option[Transaction.Hash]]("spent_finalized", O.Default(None))
```
polarker, line 45: Do we use the value `Transaction.Hash` somewhere? Or do we plan to use it later?

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -368,28 +369,29 @@ trait TransactionQueries extends StrictLogging {
...
-    }
+  def getBalanceQuery(address: Address): DBActionSR[(U256, Option[TimeStamp])] = {
+    sql"""
+      SELECT outputs.amount, outputs.lock_time
+      FROM outputs
+      LEFT JOIN inputs
+      ON outputs.key = inputs.output_ref_key
+      WHERE outputs.spent_finalized IS NULL
+      AND outputs.main_chain = true
+      AND outputs.address = $address
```
polarker, line 392: Just to confirm: will this query use `PreparedStatement`?

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
```diff
@@ -0,0 +1,152 @@
...
+      FROM inputs i
+      WHERE i.output_ref_key = o.key
+      AND o.main_chain=true
+      AND i.main_chain=true
+      AND i.block_timestamp >= $from
+      AND i.block_timestamp < $to;
+      """
+  }
+
+  def getStartEndTime()(
```
polarker, line 112: We need to add a test for this critical function, especially to test `_end.isBefore(ft)`. Just a reminder as I know there was not enough time to write tests

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
```diff
@@ -0,0 +1,152 @@
...
+ */
+
+trait FinalizerService extends SyncService.BlockFlow
+
+object FinalizerService extends StrictLogging {
+
+  // scalastyle:off magic.number
+  val finalizationDuration: Duration = Duration.ofSecondsUnsafe(6500)
+  def finalizationTime: TimeStamp    = TimeStamp.now().minusUnsafe(finalizationDuration)
+  def rangeStep: Duration            = Duration.ofHoursUnsafe(24)
```
polarker, line 46: How was the update performance after setting the timestamp range to be 24 hours?

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
```diff
@@ -0,0 +1,152 @@
...
+  private def updateOutputs(from: TimeStamp, to: TimeStamp): DBActionR[Int] = {
+    sqlu"""
+      UPDATE outputs o
+      SET spent_finalized = i.tx_hash
+      FROM inputs i
+      WHERE i.output_ref_key = o.key
+      AND o.main_chain=true
+      AND i.main_chain=true
+      AND i.block_timestamp >= $from
+      AND i.block_timestamp < $to;
```
polarker, line 108: Just to confirm: is this using `PreparedStatement`?

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
```diff
@@ -60,9 +63,13 @@ object OutputSchema extends SchemaMainChain[OutputEntity]("outputs") {
...
        lockTime,
        outputOrder,
-       txOrder)
+       txOrder,
+       spentFinalized)
         .<>((OutputEntity.apply _).tupled, OutputEntity.unapply)
   }
 
+  lazy val createNonSpentIndex: DBActionW[Int] =
+    sqlu"create index if not exists non_spent_output_idx on #${name} (main_chain,address,key, block_hash) where spent_finalized IS NULL;"
```
polarker, line 72: Nit: one extra space inside `main_chain,address,key, block_hash`

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -368,28 +369,29 @@ trait TransactionQueries extends StrictLogging {
...
+      AND inputs.block_hash IS NULL;
+    """.as[(U256, Option[TimeStamp])]
```
polarker, line 394: @simerplaha Could you help us check if this query is using the partial index `createNonSpentIndex`?

polarker (commented):
First-round is done 👍  Some comments for the future, please feel free to create issues for later improvements.

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
polarker: by adding `block_hash`, I am afraid that the index will not be used for the address query since `block_hash` is not used . @simerplaha we will need your help to investigate this part.

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
simerplaha: Think it is good idea to wrap it in some `OurDBException` so it is easier to trace DB issues in the logs or in extreme cases implement some recovery logic. I think that must have been the goal here.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
simerplaha: Yes this will use `PreparedStatement`.

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
polarker: Not to blame anything, just to raise the topic of error handling here. It's a non-trivial topic

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
simerplaha: Yes this will use `PreparedStatement`. 

```sql
(statement_name: S_1,
prepare_time:    2022-04-21 19:55:04.896275+10,
parameter_types: {bigint,bigint},
statement:       UPDATE outputs o
      SET spent_finalized = i.tx_hash
      FROM inputs i
      WHERE i.output_ref_key = o.key
      AND o.main_chain=true
      AND i.main_chain=true
      AND i.block_timestamp >= $1
      AND i.block_timestamp < $2)
```

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
polarker: Quote from Thomas's comment:
```
This new index is still used in the `getBalance` query, confirmed with an `EXPLAIN` command
```

But let's double check this as I am not sure how the query could use the index when `block_hash` is not used

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
simerplaha:
```sql
(statement_name: S_1,
prepare_time: 2022-04-21 20:00:43.256141+10,
parameter_types: {"character varying"},
statement:      SELECT outputs.amount, outputs.lock_time
      FROM outputs
      LEFT JOIN inputs
      ON outputs.key = inputs.output_ref_key
      WHERE outputs.spent_finalized IS NULL
      AND outputs.main_chain = true
      AND outputs.address = $1
      AND inputs.block_hash IS NULL)
```

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
polarker: I found the answer from the doc of Postgresql. Multi-column indexes can be used for queries that involve a subset of the columns, but the order of the columns matters here.
Quote from Postgresql: https://www.postgresql.org/docs/current/indexes-multicolumn.html
```
A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns.
```

We should probably make `address` the first column for this index? cc @tdroxler @simerplaha

polarker: but I am not sure if the index is a B-tree index or other type of index

app/src/main/scala/org/alephium/explorer/Application.scala
tdroxler: the read-write specific stuff, like db migrations and sync services

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
tdroxler: no, as in migrations we can add a new column, if we want to create an index on that column we need first to create it.

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
tdroxler: no, as in migrations we can add a new column, if we want to create an index on that column we need first to create it.

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
tdroxler: we currently recompute it when we want to find where the output is spent, when requesting a tx, I want latter to be able to reuse this instead of joining on inputs.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
tdroxler: maim-1650537499

With `EXPLAIN` we see: `using non_spent_output_idx on outputs`

polarker: As I comment in another thread, I got why the index is used even if we don't use `block_hash` here. The mechanism is more complicated than I thought, and the knowledge is very necessary for proper optimization. It turns out multi-column indexing is not that needed for output schema in this case as far as I see.

app/src/main/scala/org/alephium/explorer/Application.scala
polarker: sure, I got what is doing by diving into the code. I meant that I don't get what the function is about by just reading the function name.

app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -103,14 +106,34 @@ class Application(
...
       _ <- hashrateService.start()
+      _ <- finalizerService.start()
     } yield ()
   }
 
+  private def startReadWrite(): Future[Unit] = {
+    if (!readOnly) {
+      for {
+        _ <- dbInitializer.initialize()
+        _ <- if (readOnly) Future.successful(()) else startSyncService()
```
polarker, line 117: Here `readOnly` is false already

app/src/main/scala/org/alephium/explorer/Application.scala
polarker: The function itself is more like starting syncing services for write-only mode. We don't do anything for read-only mode.

polarker: It's nitpicking of course!

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
tdroxler: yes currently it runs every time to make it easy to use and of course we will improve the all thing in the future with a version system, I'll create a task for it.

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
tdroxler: locally all good, on aws I'm still syncing the blocks, but so far so good

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
polarker: Got it for this specific case. In general, the dependencies might be cyclic.

app/src/main/scala/org/alephium/explorer/persistence/model/AppState.scala
tdroxler: got removed in last commit

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomJdbcTypes.scala
tdroxler: removed in last commit

app/src/main/scala/org/alephium/explorer/persistence/queries/BlockDepQueries.scala
tdroxler: good point, thx :+1: it's fixed

tdroxler:
> Just a question: We've used `lazy val` at a few places here. I'm wondering if we need them at all since there is no concurrent mutation or expensive computation involved. `lazy val` generates code wrapped in `synchronized` blocks with `volatile` variable right (last time I checked for 2.12, 2.13 might be different)? So there would be a runtime cost for accessing these vals.
 
@simerplaha should we create a task to check all our lazy val?

app/src/main/scala/org/alephium/explorer/Application.scala
tdroxler: https://github.com/alephium/explorer-backend/issues/201

app/src/main/scala/org/alephium/explorer/Application.scala
tdroxler: renamed in a new commit

app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
tdroxler: https://github.com/alephium/explorer-backend/issues/202

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
tdroxler: maim-1650543338

on the image, the first explain is with the current index, I then remove it and create with `address` first. The next explain is then similar. The first `Index Cond` is just swapped, don't know if it changes something?

app/src/main/scala/org/alephium/explorer/Application.scala
tdroxler: nice catch, it's fixed

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
polarker: According to the doc, the order of the index conditions matters for indexes like B-tree. Could you check if the index type is `btree`? If so, then we'd better put `address` as the first column of the index.

Ref: https://www.postgresqltutorial.com/postgresql-indexes/postgresql-list-indexes/

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
polarker: Is it faster than 1600 update/s after increasing the range to 24 hours?

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
polarker: We will need a task for DB error handling

app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
tdroxler: maim-1650560351
 yep it's a b tree, I'll update the index.

simerplaha (approved):

simerplaha:
> should we create a task to check all our lazy val?

Sounds great :)

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
```diff
@@ -0,0 +1,164 @@
...
+    val timeRanges =
+      BlockFlowSyncService.buildTimestampRange(start, end, step)
+    foldFutures(timeRanges) {
+      case (from, to) =>
+        DBRunner
+          .run(databaseConfig)(
+            for {
+              nb <- updateOutputs(from, to)
+              _  <- updateLastFinalizedInputTime(to)
+            } yield nb
```
polarker, line 92: Let's make this action transactional?

app/src/main/scala/org/alephium/explorer/service/FinalizerService.scala
tdroxler: I don't have much logs related to outputs update as I restarted the container with a new version after the all outputs update was done, but from the  few values I can find, it's still around this 1600/2000, it's not always constant as we don't have a constant number of inputs per 24h.

polarker: okay

app/src/main/scala/org/alephium/explorer/persistence/DBRunner.scala
polarker: issue here: https://github.com/alephium/explorer-backend/issues/207

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -368,28 +365,29 @@ trait TransactionQueries extends StrictLogging {
...
+  def getBalanceQuery(address: Address): DBActionSR[(U256, Option[TimeStamp])] = {
+    sql"""
+      SELECT outputs.amount, outputs.lock_time
+      FROM outputs
+      LEFT JOIN inputs
+      ON outputs.key = inputs.output_ref_key
+      WHERE outputs.spent_finalized IS NULL
+      AND outputs.main_chain = true
+      AND outputs.address = $address
+      AND inputs.block_hash IS NULL;
```
polarker, line 389: Probably we should check `(inputs.block_hash IS NULL OR inputs.main_chain = false)`?

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -368,28 +365,29 @@ trait TransactionQueries extends StrictLogging {
...
-      .map { case (output, _) => (output.amount, output.lockTime) }
-  }
-
-  def getBalanceAction(address: Address): DBActionR[(U256, U256)] = {
-    getBalanceQuery(address).result.map { outputs =>
-      sumBalance(outputs)
-    }
+  def getBalanceQuery(address: Address): DBActionSR[(U256, Option[TimeStamp])] = {
+    sql"""
+      SELECT outputs.amount, outputs.lock_time
```
polarker, line 382: Further optimization: we could sum inside the SQL query, instead of returning the results.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -368,28 +365,29 @@ trait TransactionQueries extends StrictLogging {
...
+      FROM outputs
+      LEFT JOIN inputs
```
polarker, line 384: We will need to join based on `inputs.main_chain = true`. Otherwise, an output that is spent by an orphan block will be considered as spent in this query.

To make this query safe, I'd propose to join on only `main_chain = true` for both table, i.e. `ON (outputs.key = inputs.output_ref_key AND outputs.main_chain = true AND inputs.main_chain = true)`.

For performance, we probably need to add `main_chain` index for both inputs and outputs table.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
tdroxler: I thought about it, but can we sum two different thing? we need 2 sum, everything for the main balance and we also need to sum the one with `lock_time > now`

polarker: It should be doable in SQL. It will improve the performance when we have 1K+ UTXOs for an address

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
tdroxler: :+1: 
`main_chain` index is already there for both table

polarker: The change might affect the execution. We will need to check with the query planner, Gate.io address to make sure the performance is good.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
polarker: Issue created: https://github.com/alephium/explorer-backend/issues/209

polarker: One idea is probably to create a function to convert `outputs.lock_time` to `0 or outputs.amount` based on timestamp. Then we sum up both values.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
tdroxler: maim-1650619694
 Here are the previous and new query plan

tdroxler: I don't have all outputs updated locally so it's not yet at full speed.

app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -103,14 +106,34 @@ class Application(
...
+      } yield ()
+
+    } else {
+      Future.successful(())
+    }
+  }
+
+  private def stopReadWrite(): Future[Unit] = {
+    if (!readOnly) {
+      blockFlowSyncService.stop()
```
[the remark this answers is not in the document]
tdroxler: I'll add this to the issue related to improving services

polarker (approved):
LGTM 🚀

Issue #200, opened 2022-04-20
Improve cache


polarker:
As discussed, we could use this type of cache in many cases: https://github.com/alephium/explorer-backend/commit/e5d8dd493523dfddc490edca3d510a814c598afd

tdroxler:
Closing for now, lot's of our caches are now using the [AsyncReloadingCache](https://github.com/alephium/explorer-backend/blob/e2e/app/src/main/scala/org/alephium/explorer/cache/AsyncReloadingCache.scala)

PR #203, opened 2022-04-22
#200: Added `AsyncReloadingCache`

simerplaha:
- Towards #200
- Adds `AsyncReloadingCache`.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/cache/AsyncReloadingCache.scala
+++ b/app/src/main/scala/org/alephium/explorer/cache/AsyncReloadingCache.scala
@@ -1,0 +1,115 @@
+// Copyright 2018 The Alephium Authors
+// This file is part of the alephium project.
+//
+// The library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+package org.alephium.explorer.cache
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration.{Deadline, FiniteDuration}
+import scala.util.{Failure, Success}
+
+import com.typesafe.scalalogging.LazyLogging
+
+object AsyncReloadingCache {
+
+  /**
+    * Example:
+    * {{{
+    *    val cache =
+    *      AsyncReloadingCache(initial = "Initial value", reloadAfter = 1.second) {
+    *        currentValue: String =>
+    *          Future {
+    *            currentValue + " updated"
+    *          }
+    *       }
+    *
+    *     val value: String = cache.get()
+    * }}}
+    */
+  @inline def apply[T](initial: T, reloadAfter: FiniteDuration)(
+      loader: T => Future[T]): AsyncReloadingCache[T] =
+    new AsyncReloadingCache[T](
+      value       = initial,
+      deadline    = reloadAfter.fromNow,
+      reloading   = new AtomicBoolean(),
+      reloadAfter = reloadAfter,
+      loader      = loader
+    )
+
+}
+
+/**
+  * Cache that reloads value asynchronously returning the existing
+  * cached value while reload/re-cache is in the progress in the background.
+  *
+  * @param value       Current cached value
+  * @param deadline    Deadline until expiration
+  * @param reloading   Allows only a single thread to execute reloading
+  * @param reloadAfter How often to reload the cache or how long to keep existing cache value alive.
+  * @param loader      A function given the existing cached value returns the next value to cache.
+  * @tparam T Cached value's type
+  */
+class AsyncReloadingCache[T](@volatile private var value: T,
+                             @volatile private var deadline: Deadline,
+                             reloading: AtomicBoolean,
+                             reloadAfter: FiniteDuration,
+                             loader: T => Future[T])
+    extends LazyLogging {
+
+  /** Returns the latest cached value and invokes reload */
+  def get(): T = {
+    reload()
+    value
+  }
+
+  /**
+    * Instantly expires the current cached value and invokes reload.
+    *
+    * @note This does not always guarantee instant reload will occur immediately in this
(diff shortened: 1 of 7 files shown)


app/src/main/scala/org/alephium/explorer/cache/AsyncReloadingCache.scala
```diff
@@ -0,0 +1,115 @@
...
+/**
+  * Cache that reloads value asynchronously returning the existing
+  * cached value while reload/re-cache is in the progress in the background.
+  *
+  * @param value       Current cached value
+  * @param deadline    Deadline until expiration
+  * @param reloading   Allows only a single thread to execute reloading
+  * @param reloadAfter How often to reload the cache or how long to keep existing cache value alive.
+  * @param loader      A function given the existing cached value returns the next value to cache.
+  * @tparam T Cached value's type
```
tdroxler, line 64: I'm always amazed how you carefully write all those doc <3

tdroxler:
Question: [Some of our caches](https://github.com/alephium/explorer-backend/blob/6c4e9eaa92b2308cb5358c7efb6f5e71b67436a1/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala#L96-L103) have multiple values base on the chain-index. Should we use this cache and have `T = Map[ChainIndex, V]` or you think of adding a `MapAsyncReloadingCache`?

app/src/main/scala/org/alephium/explorer/cache/AsyncReloadingCache.scala
```diff
@@ -0,0 +1,115 @@
...
+  */
+class AsyncReloadingCache[T](@volatile private var value: T,
+                             @volatile private var deadline: Deadline,
```
polarker, line 67: I did not know about this in Scala ❤️

polarker:
LGTM, especially the atomic flag 👍  Could you try to apply this cache to our stats that could use this cache?

It's always good to see the abstraction and the actual usage in the same PR.

polarker:
It's good to merge this and open another PR to use the new cache as well. Feel free to merge if you prefer!

tdroxler:
ahah please don't use `#` in your git branch, my terminal consider it as a comment and I had to wrap with `'` to be able to checkout ^^

tdroxler:
I started to play with the cache, but one issue is for the `initialValue` which sometimes needs to be fetch from the DB. 
We maybe need a version that is blocking just for the first `get`?

polarker:
> I started to play with the cache, but one issue is for the initialValue which sometimes needs to be fetch from the DB.
We maybe need a version that is blocking just for the first get?

In our case, I don't see an initial value needed. I think we could remove the initialValue for now.

app/src/main/scala/org/alephium/explorer/cache/AsyncReloadingCache.scala
```diff
@@ -0,0 +1,115 @@
...
+          case Failure(exception) =>
+            //Fatal error: Log for debugging.
+            logger.error("Failed to reload cache", exception)
+        }
+
+        //set next deadline
+        deadline = reloadAfter.fromNow
+        //release to allow next reload
+        reloading.set(false)
+      }(ExecutionContext.parasitic)
```
tdroxler, line 113: are we sure we want this? 
https://www.scala-lang.org/api/current/scala/concurrent/ExecutionContext$$parasitic$.html
it sounds scary when reading this :sweat_smile:

polarker: Nice catch 👍  Let's use the existing execution context for now

tdroxler: > fetch and set next cache state while staying in the current
> ExecutionContext since the state update is inexpensive.

Oh ok I guess it's because of this?

polarker: I just dive into the details of this execution context. It looks great indeed. Sorry @simerplaha 

Great to know about this nice knowledge!

simerplaha:
> I started to play with the cache, but one issue is for the `initialValue` which sometimes needs to be fetch from the DB. We maybe need a version that is blocking just for the first `get`?

Yep you are right the first call has to be blocking and we can initialise the cache as `lazy` (if we think queries are expensive)? Just made a commit using this cache for `TransactionDao.cacheTxnNumber`.

> Question: [Some of our caches](https://github.com/alephium/explorer-backend/blob/6c4e9eaa92b2308cb5358c7efb6f5e71b67436a1/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala#L96-L103) have multiple values base on the chain-index. Should we use this cache and have `T = Map[ChainIndex, V]` or you think of adding a `MapAsyncReloadingCache`?

`T = Map[ChainIndex, V]` seems simplest or `Map[ChainIndex, AsyncReloadingCache[V]]`

The following example for `cachedLatestBlocks` uses your suggested `T = Map[ChainIndex, V]`

```scala
def loadLatestBlocks(): Future[Map[ChainIndex, LatestBlock]] =
  Future.traverse(chainIndexes) {
    chainIndex =>
      val latestBlocks =
        getLatestBlock(
          chainFrom = GroupIndex.unsafe(chainIndex.from.value),
          chainTo = GroupIndex.unsafe(chainIndex.to.value)
        )

      run(latestBlocks) map {
        case Some(latestBlock) =>
          chainIndex -> latestBlock

        case None =>
          ???
      }
  }.map(_.to(Map))

lazy val cachedLatestBlocks: AsyncReloadingCache[Map[ChainIndex, LatestBlock]] =
  AsyncReloadingCache(Await.result(loadLatestBlocks(), 10.seconds), 5.seconds) {
    cache =>
      loadLatestBlocks()
  }
```
There is a mutable way of doing this as well using `concurrent.TrieMap`.

simerplaha:
Ah! The use of `Await` is a little awkward. Are we ok with this or do we want another approach that doesn't use `Await`?

polarker:
> Ah! The use of Await is a little awkward. Are we ok with this or do we want another approach that doesn't use Await?

Maybe we could try to use `0` as the initial value and expire it immediately. The pseudo-code is:
```
private val cacheTxnNumber =
      AsyncReloadingCache(
        initial     = Await.result(getTotalNumberFromDB(), 5.seconds),
        reloadAfter = 5.seconds
      )(_ => getTotalNumberFromDB())
     
cacheTxnNumber.expireAndReload()
```

WDYT?

simerplaha:
Done! Seems like the best option right now. 

Updating `TransactionDao.apply` to return a `Future` would've ensured that the cache is always reloaded when the application boots-up. I will save that for another PR.

tdroxler:
mmm then for example for our `latestBlock` cache, we'll need to have as an initial value a "fake" block, but I'm fine with it, as it's really just on boot up.

polarker (approved):
LGTM 👍

PR #212, opened 2022-04-27
Minor things

simerplaha:
- Remove redundant fixme
- Remove unused `prefix: String` param from query
- Use `contains` instead of `.get(...).isEmpty`
- Use `forAll` instead of `map(...).getOrElse(true)`


tdroxler (approved):
thx!

Issue #201, opened 2022-04-21
Add a `Service` framework


tdroxler:
We should either use or build a simple framework for services so they can be started and stopped properly without having to deal with dependencies manually. Full node ref: https://github.com/alephium/alephium/blob/master/util/src/main/scala/org/alephium/util/Service.scala

tdroxler:
Let's make sure we always stop every services properly: https://github.com/alephium/explorer-backend/pull/199#discussion_r855857515

tdroxler:
Having a  service framework is now needed to be able to tackle #200 the cleaner way possible, to avoid having `await` everywhere to initialize the caches

PR #213, opened 2022-04-27
Improves PR #203 making `AsyncReloadingCache` load cache value asynchronously 

simerplaha:
# Improvements

- PR #203 [requires](https://github.com/alephium/explorer-backend/pull/203/files#diff-b16be72ab408b0bed0c9f9fe5e11270f4539a603a94c84b9300d897cde13ba23R54) `initialValue` for the cache known at compile time. This PR adds `reloadNow` function for cases where `initialValue` is unknown during compile-time and should be fetched during runtime without blocking. 
```scala
val cache: Future[AsyncReloadingCache[String]] =
  AsyncReloadingCache.reloadNow(reloadAfter = 5.seconds) {
    Future("my cached values")
  }
```
- The cache [TransactionDao.expireAndReload](https://github.com/alephium/explorer-backend/pull/203/files#diff-902043b185dee76b4d21cc0b2c9a01c46b1010286d998a2eb44128c6094b91d4R55) supplies an incorrect initial value (`0`) and fires and forget invoking `expireAndReload` hoping the cache finishes reload before the boot-up is complete. This PR shift the cache boot-up to `Application.scala` therefore guarantees that the cached value is always correct when boot-up is complete.

# Pending 

Services that start asynchronously should get initialised in `Application.start`'s `Future`. Temporarily I've added [`awaitService`](https://github.com/alephium/explorer-backend/compare/200_async_reloading_cache_2?expand=1#diff-62968abd8555cb4c8d6e190d0a3b02e74f2efd15a0670929b83ef0f13753b148R60) which blocks only once at application boot-up. This can be worked independently in another task.


app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -47,10 +48,16 @@ class Application(
     syncPeriod: Duration)(implicit system: ActorSystem, executionContext: ExecutionContext)
     extends StrictLogging {
 
+  //TOTO: Temporary placeholder for executing services returning a Future.
```
tdroxler, line 51: s/TOTO/TODO

tdroxler (approved):
Great idea, we can improve this when tackling #201

app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -47,10 +48,16 @@ class Application(
...
+  //      This will eventually be removed and the all services will get started
+  //      asynchronously via the start function.
+  def awaitService[T](f: => Future[T]): T =
+    Await.result(f, 5.seconds)
```
polarker, line 55: Nit: I am not sure it's good to make this 5 seconds implicit

polarker (approved):
LGTM ❤️

app/src/main/scala/org/alephium/explorer/Application.scala
simerplaha: Yep one of the next PRs will remove this `Await` all together. I will merge the changes made here onto the that PR and close this.

simerplaha: Kinda need to merge this. We gonna have to put up with `Await` in our code just for next the few days. 

Trying to avoid `Await` right now is a bit early and my next PR getting big with it.

PR #215, opened 2022-04-29
Resolves #209: Aggregate balance in SQL (`getBalanceAction`)

### LINKED ISSUE #209: Balance aggregation inside SQL
Ref: https://github.com/alephium/explorer-backend/pull/199#discussion_r855920679

simerplaha:
- Resolves #209
- Added `PostgresSpec` for postgres specific test-cases. 
For this PR a check that `Postgres.(epoch from now())` == `TimeStamp.now()`
- Added `SlickSugar` for queries with `exactlyOne` row.

## Benchmark 
image
`getBalanceSQL` should have been named `getBalance_ThisPR`. They are all SQL.


app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -409,9 +396,22 @@ trait TransactionQueries extends StrictLogging {
...
   }
 
-  def getBalanceAction(address: Address): DBActionR[(U256, U256)] = {
-    getBalanceQuery(address).map(sumBalance)
-  }
+  def getBalanceAction(address: Address): DBActionR[(U256, U256)] =
+    sql"""
+      SELECT sum(outputs.amount),
+             sum(CASE
+                     WHEN outputs.lock_time is NULL or outputs.lock_time < (extract(epoch from now()) * 1000) THEN 0
```
polarker, line 403: Will `epoch from now()` be called multiple times in this query? If so, we could make `now` as a query parameter. Pseudo code:
```
def getBalanceAction(address: Address, now: TimeStamp): DBActionR[(U256, U256)] =
    sql"""
      SELECT sum(outputs.amount),
             sum(CASE
                     WHEN outputs.lock_time is NULL or outputs.lock_time < $now THEN 0
```

simerplaha: Yes it will be called once. To confirm here is a test. 

```sql
-- Create over a million rows (1,078,591) in the output table
select count(*) from outputs; 

-- This query returns a single row with time value 1651467106.505805
select distinct (time) from (select extract(epoch from now()) as time from outputs) inner_query; 
```
Still gonna go with your suggestion here 👍🏼  because with time passed as parameter to the query we don't need to do timezone conversions.

polarker: Nice catch of the conversion issue 👍

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -409,9 +396,28 @@ trait TransactionQueries extends StrictLogging {
...
+                 END)
+      FROM outputs
+               LEFT JOIN inputs
+                         ON outputs.key = inputs.output_ref_key
+                             AND inputs.main_chain = true
+                             AND outputs.main_chain = true
+      WHERE outputs.spent_finalized IS NULL
+        AND outputs.address = $address
+        AND inputs.block_hash IS NULL;
+    """.as[(U256, U256)].exactlyOne
```
polarker, line 420: Thomas fixed a tricky bug for this SQL. Let's not forgot to adapt the change.

polarker (approved):
LGTM. Just one minor note

app/src/main/scala/org/alephium/explorer/util/SlickSugar.scala
```diff
@@ -0,0 +1,42 @@
...
+      * Fetch single row else fail.
+      */
+    @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
+    def exactlyOne(implicit ec: ExecutionContext): DBActionR[A] =
+      action.flatMap { rows =>
+        rows.size match {
+          case 1 => DBIO.successful(rows.head)
+          case n => DBIO.failed(new RuntimeException(s"Expected 1 result, actual $n"))
+        }
+      }
```
tdroxler, line 40: oh nice!

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -409,9 +396,28 @@ trait TransactionQueries extends StrictLogging {
...
+      lockTime = TimeStamp.now()
+    )
+
+  def getBalanceUntilLockTime(address: Address, lockTime: TimeStamp): DBActionR[(U256, U256)] =
+    sql"""
+      SELECT sum(outputs.amount),
+             sum(CASE
+                     WHEN outputs.lock_time is NULL or outputs.lock_time < ${lockTime.millis} THEN 0
+                     ELSE outputs.amount
+                 END)
```
tdroxler, line 411: Awesome, I never thought we could do 2 sums and especially using `CASE ....WHEN` like this in it.
Thx SCO (Sql Chief Officer)

tdroxler (approved):
:rocket: :rocket: :rocket:  LGTM, but just don't forget  to adapt to #218

PR #222, opened 2022-05-05
Feature/token supply

tdroxler:
We want to change the way we compute our circulating supply to match CoinMarketCap way.

Still a WIP as I need to fix some tests, but I would like to get feedback if this is really the way we want to go.


MaudSimon:
For reference, the methodology requested by CMC is in the email screenshots
https://github.com/AlephiumFoundation/Ecosystem/issues/24#issuecomment-1118363190

tdroxler:
Circulating supply is around 31M with this new method, compare to 53M with our current approach.

MaudSimon:
Also adding here that in the future, we'll need to include the burn address(es) in the list of address we should not include in the circulating supply calculation.

app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
```diff
@@ -106,30 +143,40 @@ object TokenSupplyService {
...
+        ON outputs.key = inputs.output_ref_key
+        AND inputs.main_chain = true
+        AND inputs.block_timestamp <= $at
+      WHERE outputs.block_timestamp >= ${ALPH.LaunchTimestamp} /* We don't count genesis address */
+      AND outputs.block_timestamp <= $at
+      AND outputs.main_chain = true
+      AND outputs.spent_finalized IS NULL
+      AND outputs.address NOT IN #$excludedAddresses /* We exclude the reserved wallets */
+      AND inputs.block_hash IS NULL;
+        """.as[Option[U256]].exactlyOne
```
simerplaha, line 164: Nice! I haven't tested this but since its selecting a single column I'm wondering if the `case` condition could be placed in the body somewhere so indexes get used filtering `lock_time`? In the `SELECT` we could just `sum(outputs.amount)`? 

Probably needs an index for `lock_time`.

simerplaha, line 164: Nice! I haven't tested this but since its summing a single column I'm wondering if the `case` condition (`outputs.lock_time is NULL OR outputs.lock_time <= $at THEN outputs.amount`) could be placed in the query's body so indexes get used filtering `lock_time`? Then in `SELECT` it could just be `sum(outputs.amount)`? 

Probably needs an index for `lock_time`.

simerplaha, line 164: Nice! I haven't tested this but since its summing a single column I'm wondering if the `case` condition (`outputs.lock_time is NULL OR outputs.lock_time <= $at THEN outputs.amount`) could be placed in the query's body so indexes get used filtering `lock_time`? Then in `SELECT` it could just be `sum(outputs.amount)`? 

Might need an index on `lock_time`. WDYT?

simerplaha (commented):
Nice! Looks great 👍🏼 

Tried looking for a UI on CoinMarketCap that needs complex time based queries. From this [documentation](https://support.coinmarketcap.com/hc/en-us/articles/360043396252-Supply-Circulating-Total-Max-) they seem to show only Circulating, Total, Max (Unless they need for private APIs). You reckon caching these would be helpful?

tdroxler:
> Nice! Looks great 👍🏼
> 
> Tried looking for a UI on CoinMarketCap that needs complex time based queries. From this [documentation](https://support.coinmarketcap.com/hc/en-us/articles/360043396252-Supply-Circulating-Total-Max-) they seem to show only Circulating, Total, Max (Unless they need more for private APIs). You reckon caching these would be helpful?

good point, on our side we only update those values once a day. The query to get those data isn't heavy as we just read the `token_supply` table, but still we could save lots of  db calls with caches.

app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
tdroxler: very nice catch, I started from your nice `getBalance` query but haven't thought at all of that. Will do :rocket:

tdroxler: I will add a task for the index, I tried to add a simple index on the `lock_time` column, but it's not used for the query.

app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
```diff
@@ -106,30 +143,35 @@ object TokenSupplyService {
...
+        SELECT sum(outputs.amount)
+        FROM outputs
+        LEFT JOIN inputs
+          ON outputs.key = inputs.output_ref_key
+          AND inputs.main_chain = true
+          AND inputs.block_timestamp <= $at
+        WHERE outputs.main_chain = true
+        AND outputs.block_timestamp <= $at
+        AND inputs.block_hash IS NULL;
+      """.as[U256].exactlyOne
```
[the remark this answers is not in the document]
tdroxler: oooooh of course :sweat_smile: nice catch.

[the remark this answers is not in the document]
tdroxler: It's super unlikely to happen on our mainnet since we know we'll have blocks everyday, but it could blow on a local devnet, which is unacceptable.

simerplaha, line 174: Hey Thomas, I got `NullPointerException` on empty table for running this query. I'm sure being `private` it is safe to call during runtime.

> java.lang.RuntimeException, with message: java.lang.NullPointerException: Cannot invoke "scala.math.BigDecimal.toBigInt()" because the return value of "slick.jdbc.PositionedResult.nextBigDecimal()" is null.
...

Test case
```scala
it should "empty" in {
  val tokenSupplyService =
    new TokenSupplyService.Impl(Duration.unsafe(1), databaseConfig, 1)

  run(tokenSupplyService.allUnspentTokens(TimeStamp.now())).futureValue
}
```
Should we still return `Option[U256]` here anyway so that this function is total and we don't get the `NullPointerException` (the same we saw last week)?

```scala
private def allUnspentTokensOption(at: TimeStamp): DBActionR[Option[U256]] = ...

private def allUnspentTokens(at: TimeStamp): DBActionR[U256] =
  allUnspentTokensOption(at).map(_.getOrElse(U256.Zero))
```

tdroxler: ooooh of course :sweat_smile:  nice catch

tdroxler:
I need to update the computation: I correctly don't count the genesis amout, but then I forgot to exclude the unlocked amount of the genesis addresses.

h0ngcha0 (commented):
LGTM, great work, just a few questions for myself :)

app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
```diff
@@ -27,16 +27,39 @@ import slick.dbio.DBIOAction
...
+ *   Burn
+ *   Escrow
+ *   Private sale investors
+ *   Marketing operations
+ *   Treasury
+ *   Ecosystem incentives
+ *   Team/advisor/contractors
+ *   Team-conrolled assets
+ *
+ * Most of those wallets are included in our genesis blocks, so we don't take those blocks when computing the circulating supply,
```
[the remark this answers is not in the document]
MaudSimon: image
@h0ngcha0

tdroxler:
With the latest commit I now correctly always exclude the genesis addresses from any computation, circulating tokens with this is now around 15M

app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
tdroxler: thx :+1: I fixed in a new commit.

polarker:
I don't have the time to review it right now. If we are in a hurry to merge it, let's get @h0ngcha0 's approval instead of mine.

app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
```diff
@@ -18,24 +18,73 @@ package org.alephium.explorer.persistence
...
+  def getVersion()(implicit ec: ExecutionContext): DBActionR[Int] = {
+    sql"""
+      SELECT value FROM app_state where key = 'migrations_version'
+    """.as[ByteString].headOption.map {
+      case None => 0
+      case Some(bytes) =>
+        deserialize[Int](bytes) match {
+          case Left(error) =>
+            logger.error(s"Invalid migration version, closing app. $error")
+            System.exit(1)
```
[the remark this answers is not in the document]
tdroxler: fixed

app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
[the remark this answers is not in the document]
h0ngcha0: Thanks a lot for the explanation @MaudSimon  :)

h0ngcha0 (approved):
LGTM, awesome work :rocket: :rocket: :rocket:

PR #227, opened 2022-05-10
Updated `DBInitializer` to be stateless

simerplaha:
- First PR for a series of other PRs working towards #201
- Updates `DBInitializer` to be stateless

The issue #201 can be implemented framework-less if majority of the code is made stateless. Previously statefulness was required because of H2 dependency, that got resolved by Thomas in #163. 

Being simple objects a plain function our services would be just a function (`syncOnce`) that's executed periodically by a `Scheduler` (implemented in one of the commits in the screenshot) making the code much simpler.

The following types are the only global parameters required by 90% of the code and there is no mutation.
- `ExecutionContext` 
- `DatabaseConfig[PostgresProfile]`
- `GroupConfig` (for some)

Mutation occurs only in caches which is encapsulated and managed in a dedicated class.

I've got the following commits that will be submitted in individual PRs so its easier to review. 

image

### DIFF
--- a/app/src/test/scala/org/alephium/explorer/persistence/DatabaseFixture.scala
+++ b/app/src/test/scala/org/alephium/explorer/persistence/DatabaseFixture.scala
@@ -16,7 +16,7 @@
 
 package org.alephium.explorer.persistence
 
-import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor}
+import scala.concurrent.{Await, ExecutionContext}
 import scala.jdk.CollectionConverters._
 
 import com.typesafe.config.{ConfigFactory, ConfigValueFactory}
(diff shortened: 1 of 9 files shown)


app/src/test/scala/org/alephium/explorer/persistence/DatabaseFixture.scala
```diff
@@ -40,10 +40,10 @@ object DatabaseFixture {
   def createDatabaseConfig(): DatabaseConfig[PostgresProfile] =
     DatabaseConfig.forConfig[PostgresProfile]("db", config)
 
-  def dropCreateTables(databaseConfig: DatabaseConfig[PostgresProfile]) = {
-    val dbInitializer: DBInitializer = new DBInitializer(databaseConfig)(ExecutionContext.global)
+  def dropCreateTables()(implicit databaseConfig: DatabaseConfig[PostgresProfile]) = {
+    implicit val ec: ExecutionContextExecutor = ExecutionContext.global
```
tdroxler, line 44: why can't it just be an `ExecutionContext` ? it's just "cleaner" to have less import :stuck_out_tongue_closed_eyes:

tdroxler (approved):
Awesome :rocket:  let's simplify that code base!!!

PR #228, opened 2022-05-10
Stateless `UnconfirmedTxDao` & `HealthCheckDao`

simerplaha:
- Second PR towards issue #201
- Updated `UnconfirmedTxDao` & `HealthCheckDao` to be stateless
- Build on top of PR #227

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/Application.scala
+++ b/app/src/main/scala/org/alephium/explorer/Application.scala
@@ -50,7 +50,6 @@ class Application(host: String,
   val blockDao: BlockDao                = BlockDao(groupNum, databaseConfig)
   val transactionDao: TransactionDao    = TransactionDao(databaseConfig)
   val utransactionDao: UnconfirmedTxDao = UnconfirmedTxDao(databaseConfig)
-  val healthCheckDao: HealthCheckDao    = HealthCheckDao(databaseConfig)
 
   //Services
   val blockFlowClient: BlockFlowClient =
@@ -130,7 +129,7 @@ class Application(host: String,
 
   def start: Future[Unit] = {
     for {
-      _       <- healthCheckDao.healthCheck()
+      _       <- HealthCheckDao.healthCheck()
       _       <- startTasksForReadWriteApp()
       binding <- Http().newServerAt(host, port).bindFlow(server.route)
     } yield {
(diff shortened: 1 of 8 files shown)


app/src/main/scala/org/alephium/explorer/persistence/dao/HealthCheckDao.scala
```diff
@@ -18,30 +18,20 @@ package org.alephium.explorer.persistence.dao
...
 
-import com.typesafe.scalalogging.StrictLogging
 import slick.basic.DatabaseConfig
 import slick.jdbc.PostgresProfile
 import slick.jdbc.meta.MTable
 
-import org.alephium.explorer.persistence.DBRunner
+import org.alephium.explorer.persistence.DBRunner._
 
 trait HealthCheckDao {
```
tdroxler, line 27: we can remove that trait right?

app/src/main/scala/org/alephium/explorer/Application.scala
```diff
@@ -59,7 +58,7 @@ class Application(host: String,
     BlockFlowSyncService(groupNum = groupNum, syncPeriod = syncPeriod, blockFlowClient, blockDao)
 
   val mempoolSyncService: MempoolSyncService =
-    MempoolSyncService(syncPeriod = syncPeriod, blockFlowClient, utransactionDao)
+    MempoolSyncService(syncPeriod = syncPeriod, blockFlowClient, UnconfirmedTxDao)
```
tdroxler, line 61: let's remove the `UnconfirmedTxDao` trait and use directly the object inside `MempoolSynService` without pasing it as arg no?

simerplaha: Yep. This will be removed when `MempoolSyncService` turns into `object MempoolSyncService` so this constructor disappears entirely. Can we wait for this change till then? Each PR will be bare minimum changes required for the targeted class.

tdroxler: sure

tdroxler (approved):

app/src/main/scala/org/alephium/explorer/persistence/dao/HealthCheckDao.scala
simerplaha: Yes we can. Done :)

PR #231, opened 2022-05-10
Fix wrong supply due to one genesis address

tdroxler:
One of the address of the genesis file is not part of the reserved
wallets, we thus remove it from the reserved address, but we can't
filter out outputs from the genesis blocks anymore. We can only rely on the
`NOT IN` clause to filter the reserved genesis addresses.


simerplaha (approved):
🚀 🚀

PR #232, opened 2022-05-12
#201 Added `Scheduler`

simerplaha:
- Towards #201 
- Added a simple `Scheduler`

@tdroxler mentioned we have a use-case where "it needs to be ran once a day, just after midnight". See `scheduleDailyAt` below. This `Scheduler` is not being used right now but will be after others are made stateless.

```scala
val scheduler = Scheduler("My scheduler")

//once only
val once =
  scheduler.scheduleOnce(1.second) {
    Future {
      "Do something expensive"
    }
  }

//infinitely every one second
val loop =
  scheduler.scheduleLoop(1.second) {
    Future {
      "Do something expensive"
    }
  }

//infinitely every one second
val initAndLoop =
  scheduler.scheduleLoopFlatMap(1.second) {
    //initialisation code only called once
    Future {
      "Some initialisation operation"
    }
  } {
    // the initialise value processed above
    initialisedValue: String =>
      Future {
        "Some expensive operation"
      }
  }

//Schedule once a day at 1AM
val daily =
  scheduler.scheduleDailyAt(LocalTime.of(1, 0)) {
    Future {
      "Some expensive operation"
    }
  }
```


polarker:
Looks like a charm ❤️ Will try to read and learn from it

app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
```diff
@@ -0,0 +1,187 @@
...
+      name       = name,
+      timer      = new Timer(name, isDaemon),
+      terminated = false
+    )
+
+  /**
+    * Calculate time left to schedule for the input [[LocalTime]].
+    * If the time is in the past then the schedule for tomorrow.
+    *
+    * TimeZone used is local machine's default time-zone
```
tdroxler, line 51: mmm I think we should try to work with UTC time only, for example for the `scheduleDailyAt` I will set it up at 1am, because I really want it to run just when a UTC day is finished. 

As I don't know what's the time-zone of the AWS machine for prod it's impossible :/

tdroxler (commented):
Did a first round and it looks super nice, just this UTC time we need to discuss.

app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
simerplaha: Ah yes good idea. For this use-case made the following changes

- Added `scheduleDailyAtUTC(utcTime)`.
- Changed the function to use `LocalDateTime` instead of `LocalTime` to specify the **day** to start scheduling.

WDYT?
 
The highlighted function `scheduleTime` calculates `FiniteDuration` left until next schedule right, so here UTC or Local wouldn't make a difference to the final output. Prefer using local system time here because it's easier when debugging & testing, easy to visually compare time against the time on my machine.

simerplaha: As requested on Slack I've switched
- `LocalTime` to `OffsetTime`
- `LocalDateTime` to `ZonedDateTime`

app/src/test/scala/org/alephium/explorer/util/SchedulerSpec.scala
```diff
@@ -15,63 +15,83 @@
...
+class SchedulerSpec
+    extends AnyWordSpec
+    with ScalaCheckDrivenPropertyChecks
+    with Matchers
+    with Eventually
+    with ScalaFutures {
 
   implicit val executionContext: ExecutionContext = ExecutionContext.global
 
+  def zoneIds(): Iterable[ZoneId] =
```
tdroxler, line 45: why not a `val`?

app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
```diff
@@ -144,28 +142,12 @@ class Scheduler private (name: String, timer: Timer, @volatile private var termi
...
   /**
     * Schedules daily, starting from today.
-    *
-    * @param atUTC Should be in UTC.
     */
   @SuppressWarnings(Array("org.wartremover.warts.Overloading"))
-  def scheduleDailyAtUTC[T](atUTC: LocalTime)(block: => Future[T])(
-      implicit ec: ExecutionContext): Unit =
-    scheduleDailyAt(TimeUtil.toLocalDateTimeNow(atUTC))(block)
+  def scheduleDailyAt[T](at: OffsetTime)(block: => Future[T])(implicit ec: ExecutionContext): Unit =
```
tdroxler, line 149: niiiiice, it clean a bit the code

app/src/main/scala/org/alephium/explorer/util/TimeUtil.scala
```diff
@@ -0,0 +1,27 @@
...
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+package org.alephium.explorer.util
+
+import java.time._
+
+object TimeUtil {
```
tdroxler, line 21: great, I have few functions to move here too :rocket:

app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
```diff
@@ -0,0 +1,190 @@
...
+    val timeLeft = Duration.between(ZonedDateTime.now(scheduleAt.getZone), scheduleAt)
+
+    if (timeLeft.isNegative) { //time is in the past, schedule for tomorrow.
+      val daysBehind = Math.abs(timeLeft.toDays) + 1
+      logger.trace(s"Scheduled time is $daysBehind.days behind.")
+      scheduleTime(scheduleAt.plusDays(daysBehind))
+    } else { //time is in the future. Good!
+      val nextSchedule = timeLeft.toNanos.nanos
+      //calculate first schedule using today's date.
+      logger.debug(s"Scheduled task after ${nextSchedule.toSeconds}.seconds")
```
tdroxler, line 68: might be nice in this log and the one above to get the `name` of the scheduler?

tdroxler (approved):
one very minor about logging, otherwise LGTM :rocket: I gave it a try and it's super cool!

app/src/main/scala/org/alephium/explorer/util/TimeUtil.scala
polarker: I will add some black magic here later to improve the performance of `TimeStamp.now` when it becomes a bottleneck :)

PR #235, opened 2022-05-18
Stateless `HashrateService` & `TokenSupplyService`

simerplaha:
- Towards issue #201
- Stateless `HashrateService` & `TokenSupplyService`
- Both services use the `Scheduler` from #232

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
+++ b/app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
@@ -53,19 +53,19 @@ object Scheduler extends StrictLogging {
     * @return Time left until next schedule.
     */
   @tailrec
-  def scheduleTime(scheduleAt: ZonedDateTime, schedulerName: String): FiniteDuration = {
+  def scheduleTime(scheduleAt: ZonedDateTime, id: String): FiniteDuration = {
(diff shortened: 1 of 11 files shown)


app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
```diff
@@ -117,7 +117,9 @@ class Scheduler private (name: String, timer: Timer, @volatile private var termi
       case Failure(exception) =>
         //Log the failure.
         logger.error(s"${logId(taskId)}: Failed executing task", exception)
-        scheduleLoop(taskId, loopInterval, loopInterval)(block)
+        if (!terminated) {
+          scheduleLoop(taskId, loopInterval, loopInterval)(block)
+        }
```
tdroxler, line 122: git extremist here: This change doesn't seems to be related to that commit message :)

tdroxler (approved):
LGTM :rocket:

app/src/main/scala/org/alephium/explorer/util/Scheduler.scala
simerplaha: Ah my bad. I was being lazy. Will make sure next time :)

PR #238, opened 2022-05-19
fix: `using.resource` should be initialised immediately

simerplaha:


tdroxler (approved):

tdroxler:
this sound's like the bug complicated to find ^^

simerplaha:
Yep 😆

Issue #206, opened 2022-04-22
Double check if all transactional actions are using `.transactionally`


polarker:

tdroxler:
one of them to check was discussed here:
https://github.com/alephium/explorer-backend/pull/275#discussion_r903248190

tdroxler:
I did check and I think we are good. I might have miss some, but our queries are stable since a long time, so  we can close the issue for now.

Issue #204, opened 2022-04-22
Dedicated `ExecutionContext` for services & IO


simerplaha:
Currently all services (all sync and reads) share the same `ExecutionContext` supplied by the `ActorSystem`. Since we lean more towards read heavy workloads there might be a need for dedicated `ExecutionContext` favouring reads. 

Slick manages its own `ExecutionContext` for IO which is great but double check its concurrency settings are suited for IO bounded tasks.

Edit (23/05/2022): See [comment](https://github.com/alephium/explorer-backend/pull/239#discussion_r879085519)

simerplaha:
See [comment](https://github.com/alephium/explorer-backend/pull/255#issuecomment-1149638426): `ActorSystem` is unnecessarily being started in the `Main` which is only used by akka-http. Explorer itself does not use Akka.

simerplaha:
Also see [comment](https://github.com/alephium/explorer-backend/pull/380#discussion_r1016191005).

PR #239, opened 2022-05-20
Feature/tx history

### LINKED ISSUE #211: Create endpoint to get the number of TX per day (history)
The goal being to produce something like : 

image

This would be the chart produced in the list compiled in https://github.com/alephium/explorer/issues/32#issue-1062316734

tdroxler:
resolves #211 

I still having the feeling I could improve a lot this PR, but I feel I need some feedback too, otherwise I could pass a whole week on it :stuck_out_tongue_closed_eyes: 

The new endpoint is:

`/charts/transactions-count?fromTs=1234&toTs=12345&interval-type=daily`

The `interval-type` cand be `hourly` or `daily`.
There is also a non-mandatory query parameter: `per-chain=true` (deault is `false`), which return the list of count per chain if set to true and nothing otherwise, see below the response:

```
[
  {
    "timestamp": 12345,
    "total": 1234,
    "values": [ // OPTIONAL
      {
        "chainFrom": 0,
        "chainTo": 0,
        "value": 1234
      }
    ]
  }
]
```

Let's discuss the format @lf94 , I will anyway wait your input before merging/releasing that PR.


app/src/main/scala/org/alephium/explorer/service/TransactionHistoryService.scala
```diff
@@ -0,0 +1,233 @@
...
+    run(findLatestHistoryTimestamp(intervalType)).flatMap { histTsOpt =>
+      val start = histTsOpt
+        .map { histTs =>
+          stepBack(histTs, intervalType)
+        }
+        .getOrElse(ALPH.LaunchTimestamp)
+
+      val ranges = getTimeRanges(start, latestTxTs, intervalType)
+
+      foldFutures(ranges) {
```
simerplaha, line 120: For more parallelism do we wanna consider using `Future.traverse(ranges)`? Assuming `ranges` contains dates that are not overlapping.

tdroxler: actually when you sync the first time or you didn't sync for several days, it fails with a `traverse` because the db can't handle that much connections and inserts at the same time.

simerplaha: Ah yep! Makes sense! 👍🏼  This probably should be looked into by #204: motivation being that we shouldn't have to reduce in-code performance (CPU based operations) to satisfy performance deficiency is database or any IO, this should be handled within thread-pools and connection-pools.

tdroxler: Fully agree :+1:

simerplaha (commented):
Looks great 🚀 . Gave it another look today.

LeeAlephium:
I would prefer something a little more self-descriptive but same structure:

```
[
  {
    "timestamp": 12345,
    "totalAllChains": 1234,
    "totalPerChain": [ // OPTIONAL
      {
        "chainFrom": 0,
        "chainTo": 0,
        "value": 1234
      }
    ]
  }
]
```

tdroxler:
all good with a more descriptive version :+1: 

> I'm actually thinking maybe it's not great to aggregate by default... This is going to cause a per-chain / aggregate for almost all operations for explorer. I think it's better if we let the consumer determine what to do?

@LeeAlephium so you mean when we call `per-chain = true` we only return `totalPerChain` and not `totalAllChains`? or even we just get rid of `per-chain` and always return the `totalPerChain` and let user sum-up?

tdroxler:
we got an internal discussion and we think we could even go a more explicit and cleaner way:

two endpoints with each of them their model, so there's no confusion:

`/charts/transactions-count`
```
[
  {
    "timestamp": 12345,
    "totalAllChains": 1234,
  }
]
```

and

`/charts/transactions-count-per-chain`
```
[
  {
    "timestamp": 12345,
    "totalPerChain": [
      {
        "chainFrom": 0,
        "chainTo": 0,
        "value": 1234
      }
    ]
  }
]
```

app/src/main/scala/org/alephium/explorer/persistence/schema/TransactionHistorySchema.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.alephium.explorer.persistence.schema.CustomJdbcTypes._
+import org.alephium.util.TimeStamp
+
+object TransactionHistorySchema extends Schema[TransactionHistoryEntity]("transactions_history") {
+
+  class TransactionsHistories(tag: Tag) extends Table[TransactionHistoryEntity](tag, name) {
+    def timestamp: Rep[TimeStamp]       = column[TimeStamp]("timestamp")
+    def chainFrom: Rep[GroupIndex]      = column[GroupIndex]("chain_from")
+    def chainTo: Rep[GroupIndex]        = column[GroupIndex]("chain_to")
+    def value: Rep[Long]                = column[Long]("value")
```
polarker, line 33: maybe rename this as `count` as well

simerplaha (approved):
🚀 

Just to confirm for issue #206 - `update*` queries in `TransactionHistoryService` do not need to be transactional right?

tdroxler:
> Just to confirm for issue #206 - `update*` queries in `TransactionHistoryService` do not need to be transactional right?

oh good point, I think yes they need to be, we need to make sure that for a given time range we update all chains or none, otherwise the next time we restart we won't recover those missing values for a given time range.
Let me push a new commit

tdroxler:
@mvaivre @nop33 @LeeAlephium in [this commit](https://github.com/alephium/explorer-backend/pull/239/commits/273099328539954d4546a0db89670b1308d482a3)
We change a bit our models to be explicit, to give an idea it's something like:
```
/heights => [{"chainFrom": 1, "chainTo": 2, "height": 100, "value": 100}]
/block-time => [{"chainFrom": 1, "chainTo": 2, "duration": 100, "value": 100}]
/tx-counts => [{"chainFrom": 1, "chainTo": 2, "count": 100}]
```
For the two first endpoint I kept for now the `value` field to avoid breaking the front-end, but it will be removed latter once we migrate to those new models.

tdroxler:
@mvaivre @nop33 @LeeAlephium last point, when using `hourly` time-interval, the time-range will be limited at 30 days, no restriction when using `daily`.

PR #241, opened 2022-05-23
Stateless `SanityChecker`

simerplaha:
- Towards #201
- Stacked PR of previous un-merged PR #240
- Stateless `SanityChecker`

`SanityChecker` has mutable vars `running`, `i`, `totalNbOfBlocks` which from a quick look can be removed and implemented immutably. Is this something we want done?

# Pending stateless services

Two more left - `BlockFlowSyncService` and `MempoolSyncService`.


tdroxler:
>SanityChecker has mutable vars running, i, totalNbOfBlocks which from a quick look can be removed and implemented immutably. Is this something we want done?

Let's create an issue for that? unless you think it's critical?

simerplaha:
> Let's create an issue for that? unless you think it's critical?

Should be low priority. Created #242 :)

tdroxler (approved):

PR #250, opened 2022-05-26
Stateless sync service

simerplaha:
- Towards #201
- Updated the final two services (`BlockFlowSyncService` & `MempoolSyncService`) to be stateless.
- Added `scheduleLoopConditional` to `Scheduler`. Used by `BlockFlowSyncService`. Already existing `scheduleLoopFlatMap` didn't cover the use-case. Do we want to keep `scheduleLoopFlatMap` around?
- Added `LazyVal` which behaves exactly scala's `lazy` keyword but provides state information if the value was initialised or not. This was needed for [this](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/ApplicationSpec.scala#L426) test-case ("scheduler should  not be initialised in read-only mode") but as discussed on Slack, tests in `ApplicationSpec` are temporarily ignored so `LazyVal` is not being used right now, do we want to keep `LazyVal` around for any future use-cases? 
```scala
implicit val scheduler: LazyVal[Scheduler] =
  LazyVal(Scheduler("Sync scheduler"))
``` 
- Removed `trait SyncService`


tdroxler:
> Do we want to keep scheduleLoopFlatMap around 

Let's remove it if we don't use it, less code to maintain. If you remove just that fonction in a nicely named commit, we can easily revert it back latter. 

> do we want to keep LazyVal around for any future use-cases? 

I would prefer not having to much wrappers and abstractions if we can avoid them, let's KISS it!
But if you need it during the big current refactoring, we can remove it latter.

tdroxler (approved):

PR #254, opened 2022-05-31
Remove unused and duplicate function in various places

tdroxler:
We kept lots of functions that were written in `slick` just to validate
that the new corresponding SQL functions were correct.

I think we ran everything in prod for long enough to validate it's correct.

This was also getting annoying to maintain as soon as we are adding a column somewhere.


tdroxler:
image
that's the kind of picture I like to see :)

simerplaha (approved):
Awesome!!! 🚀 🚀 🚀

Issue #246, opened 2022-05-24
Handle system termination (`sys.exit`) gracefully


simerplaha:
Currently `sys.exit(...)` in invoked at a few places making those function untestable. 

System termination should be centralised and should happen after all services are closed properly instead of abrupt jvm termination. 

Also check for alternate solution to system termination. Is halting sync services and alerting admins (email or whatever) of invalid state enough instead of jvm termination?

polarker:
We will need to sync with Benoit. As far as I see, we could
* for fatal errors, we log them, terminate the system, and let docker restart the application automatically
* for non-fatal errors, logging it properly should be enough

tdroxler:
Had a call yesterday with Benoît and I created #243 from ops side it's always better to restart on fatal-error instead of trying to recover.

Issue #207, opened 2022-04-22
Improve DB error handling


polarker:

simerplaha:
# Checkpoint documentation

Switched worked on higher priority tasks. Noted the following list of things TODO so far. Will continue after resolving higher priority tasks.

- Are logs being logged to a file? Since there is no file appender in `logback.xml`, I’m assuming logs are piped into a file?
- Are transactions that are rollbacked logged?
- Check that when application starts up and a query is executed, that failed queries are logged.
- How much is sync logging? Are sync’s logging to the same file?
- Are failed web requests logged?
- Is query timeout logged?
- Check that logs are not unnecessarily overflowed.

polarker:
Here is the rollback configuration for full node: https://github.com/alephium/alephium/blob/master/flow/src/main/resources/logback.xml

PR #255, opened 2022-06-07
Resolves #201: Add a Service framework

### LINKED ISSUE #201: Add a `Service` framework

simerplaha:
# Overview

- Resolves #201
- Partially towards #246
- Partially towards #207

# Configuration

- Moved all `syncPeriod` configuration to `application.conf` file.
- Added `ApplicationConfig`: A typed `application.conf` file.
- Added `ExplorerConfig`: A validated config for starting `Explorer` ensuring `Explorer` does not get started with incorrect configurations.

# Updates

- Renamed `Application` to `Explorer`.
- Updated `Explorer` to boot-up executing a sequence of functions.
- `Explorer` boots in either one of two states
    - `ExplorerState.ReadOnly`
    - `ExplorerState.ReadWrite`
- Removed calls to `sys.exit` from `Application` (Towards #246). On boot-up error, server is not started so JVM automatically stops.
- Added `ExplorerError` for typed error/exceptions (Towards #207).
- Added `AsyncCloseable` for managing resource termination (one or many) when there is a failure. See `FutureUtil.managed` &  test-cases.

# Testing

- Added dependency on [ScalaMock](http://scalamock.org/). I know mocking is taboo, but I think it's needed here. Helps writing readable test-cases. Something like [this](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/service/MempoolSyncServiceSpec.scala#L81) and [this](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/web/InfosServerSpec.scala#L156) is using `???` which can made much simpler with mocking. Also helps with performance testing/debugging.
- Added multiple helper ScalaCheck test-data generators. First step towards splitting `Generators` into multiple smaller files.
    - `GenCommon` - Generators for types supplied by libraries outside Alephium eg: java or scala packages
    - `GenCoreApi` - Generators for types supplied by Core `org.alephium.api` package
    - `GenCoreProtocol` - Generators for types supplied by Core `org.alephium.protocol` package

# Pending

- `ExplorerConfig`'s validations methods might be missing some validation checks. Eg is `0` a valid group-number?
- `ApplicationSpec`'s existing test-cases still need to be ported. Created issue #256 for this task. 
- A question marked under `TODO` in `Explorer.getBlockFlowPeers`

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/AppServer.scala
+++ b/app/src/main/scala/org/alephium/explorer/AppServer.scala
@@ -38,9 +38,9 @@ object AppServer {
                transactionCache: TransactionCache,
                groupSetting: GroupSetting): Route = {
 
-    val blockServer                = new BlockServer(BlockService)
+    val blockServer                = new BlockServer()
     val addressServer              = new AddressServer(TransactionService)
-    val transactionServer          = new TransactionServer(TransactionService)
+    val transactionServer          = new TransactionServer()
     val infosServer                = new InfosServer(TokenSupplyService, BlockService, TransactionService)
     val utilsServer: UtilsServer   = new UtilsServer()
     val chartsServer: ChartsServer = new ChartsServer()
--- a/app/src/main/scala/org/alephium/explorer/AppServer.scala
+++ b/app/src/main/scala/org/alephium/explorer/AppServer.scala
@@ -21,7 +21,6 @@ import scala.concurrent.ExecutionContext
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
 import ch.megard.akka.http.cors.scaladsl.CorsDirectives._
-import com.typesafe.scalalogging.StrictLogging
 import slick.basic.DatabaseConfig
 import slick.jdbc.PostgresProfile
 
@@ -30,27 +29,22 @@ import org.alephium.explorer.service._
 import org.alephium.explorer.web._
 
 // scalastyle:off magic.number
-class AppServer(blockService: BlockService,
-                transactionService: TransactionService,
-                tokenSupplyService: TokenSupplyService)(implicit executionContext: ExecutionContext,
-                                                        dc: DatabaseConfig[PostgresProfile],
-                                                        blockFlowClient: BlockFlowClient,
-                                                        blockCache: BlockCache,
-                                                        transactionCache: TransactionCache,
-                                                        groupSetting: GroupSetting)
-    extends StrictLogging {
+object AppServer {
 
-  val blockServer: BlockServer = new BlockServer(blockService)
-  val addressServer: AddressServer =
-    new AddressServer(transactionService)
-  val transactionServer: TransactionServer =
-    new TransactionServer(transactionService)
-  val infosServer: InfosServer =
-    new InfosServer(tokenSupplyService, blockService, transactionService)
-  val utilsServer: UtilsServer   = new UtilsServer()
-  val chartsServer: ChartsServer = new ChartsServer()
+  def routes()(implicit ec: ExecutionContext,
+               dc: DatabaseConfig[PostgresProfile],
+               blockFlowClient: BlockFlowClient,
+               blockCache: BlockCache,
+               transactionCache: TransactionCache,
+               groupSetting: GroupSetting): Route = {
 
-  val route: Route =
+    val blockServer                = new BlockServer(BlockService)
+    val addressServer              = new AddressServer(TransactionService)
+    val transactionServer          = new TransactionServer(TransactionService)
+    val infosServer                = new InfosServer(TokenSupplyService, BlockService, TransactionService)
+    val utilsServer: UtilsServer   = new UtilsServer()
+    val chartsServer: ChartsServer = new ChartsServer()
+
     cors()(
       blockServer.route ~
         addressServer.route ~
@@ -60,4 +54,5 @@ class AppServer(blockService: BlockService,
         utilsServer.route ~
         DocumentationServer.route ~
         Metrics.route)
+  }
 }
--- a/app/src/main/scala/org/alephium/explorer/Explorer.scala
+++ b/app/src/main/scala/org/alephium/explorer/Explorer.scala
@@ -250,9 +250,10 @@ object Explorer extends StrictLogging {
       blockFlowClient: BlockFlowClient): Future[Seq[Uri]] =
     if (directCliqueAccess) {
       blockFlowClient.fetchSelfClique() flatMap {
+        case Right(selfClique) if selfClique.nodes.isEmpty =>
+          Future.failed(PeersNotFound(blockFlowUri))
+
         case Right(selfClique) =>
-          //TODO: What happens when peers are empty?
-          //      Should Application even be allowed to start if there were no peers?
           val peers = urisFromPeers(selfClique.nodes.toSeq)
           logger.debug(s"Syncing with clique peers: $peers")
           Future.successful(peers)
--- a/app/src/main/scala/org/alephium/explorer/Explorer.scala
+++ b/app/src/main/scala/org/alephium/explorer/Explorer.scala
@@ -54,7 +54,7 @@ object Explorer extends StrictLogging {
       case Success(config)    => Explorer(config)
     }
 
-  /** Start Explorer from `application.conf` */
+  /** Start Explorer from parsed/loaded `application.conf` */
   def apply(config: Config)(implicit ec: ExecutionContext): Future[ExplorerState] =
     ApplicationConfig(config) match {
       case Failure(exception)         => Future.failed(exception)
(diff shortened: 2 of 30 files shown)


tdroxler:
>I know mocking is taboo  

Ahaha indeed I always had bad experiences with scala mock, it's full of nice features on the paper, but as soon as you try to use it heavily, it's a nightmare IMO. But I'm ok to give it a try again, maybe we should have some rule to say that we only use it to mock unused stuff to replace the `???`

simerplaha:
> I always had bad experiences with scala mock, it's full of nice features on the paper, but as soon as you try to use it heavily, it's a nightmare IMO.

Oh that's no fun. Was it an older version? Newer version should work well.

> maybe we should have some rule to say that we only use it to mock unused stuff to replace the `???`

Yep totally! The Gods of programming have always said _"Mocking mostly implies the design smells!"_. But at rare times I've found mock being super handy for writing complete tests given the situation. In addition to replacing `???`, it be nice if we were ok with mocking if:

- If it helps writing complete tests covering all cases (high coverage) of a function being tested, it's ok.
- If it helps writing more readable & maintainable tests without spinning my own little framework around each test-case mutating things or using `???`, it's ok.

Will definitely try to minimise the use of mock so each test covers more lines of code. Can only see less than 10% of our tests using mock.

simerplaha:
Issue #258 reports timeout during migration. This PR uses `await` in `Main.scala` for which the timeout should be configurable. Do we prefer the timeout be configured via `application.conf` file or via command line?

polarker:
> Do we prefer the timeout be configured via application.conf file or via command line?

Is it possible to chain all of those bootstrap Futures and make them run asynchronously without `await`?

simerplaha:
> Is it possible to chain all of those bootstrap Futures and make them run asynchronously without `await`?

Done! Relying on `ActorSystem` to handle non-daemon startup by initialising it first, which is kinda unnecessary since we only use Akka for the server but it gets rid of the `Await`. 

Will have the properly resolved in #204 with custom `ThreadPool`.

app/src/main/scala/org/alephium/explorer/config/ExplorerConfig.scala
```diff
@@ -0,0 +1,159 @@
...
+        hashRateServiceSyncPeriod       = hashRateServiceSyncPeriod,
+        finalizerServiceSyncPeriod      = finalizerServiceSyncPeriod,
+        transactionHistoryServicePeriod = transactionHistoryServicePeriod
+      )
+
+  def validateGroupNum(groupNum: Int): Try[Int] =
+    if (groupNum < 0) {
+      Failure(InvalidGroupNumber(groupNum))
+    } else {
+      //Is 0 a valid groupNum? Is there a max limit?
```
tdroxler, line 80: yes 0 is a valid group. We could check latter what we use to validate group in the node.

app/src/main/scala/org/alephium/explorer/util/FutureUtil.scala
```diff
@@ -16,18 +16,78 @@
...
+    *  }
+    * }}}
+    *
+    * Multiple resources can be initialised with embedding. If the body of any
+    * one of the resource fails, all parent resources are closed ensuring no leaks.
+    * {{{
+    *   managed(ActorSystem("")) { resource1 =>
+    *     managed(Scheduler("")) { resource2 =>
+    *       managed(someResource()) { resource3 =>
+    *          Future(???) //If this Future fails, all resources above are closed
```
tdroxler, line 62: nice!

tdroxler (commented):
>ApplicationSpec's existing test-cases still need to be ported. 

yep we need them back soon, as we are changing a lot of things here.

I did a first round, looks super promising. I'll now do some manual testing.

tdroxler:
Manual testing went all good :+1:

app/src/main/scala/org/alephium/explorer/AppServer.scala
```diff
@@ -30,27 +29,22 @@ import org.alephium.explorer.service._
...
+object AppServer {
+
+  def routes()(implicit ec: ExecutionContext,
+               dc: DatabaseConfig[PostgresProfile],
+               blockFlowClient: BlockFlowClient,
+               blockCache: BlockCache,
+               transactionCache: TransactionCache,
+               groupSetting: GroupSetting): Route = {
+
+    val blockServer                = new BlockServer(BlockService)
```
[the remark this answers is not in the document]
simerplaha: Oh this these server classes weren't updated in this PR. Just the `AppServer`. But you are right, might as well remove unneeded parameters now 👍🏼

app/src/main/scala/org/alephium/explorer/Main.scala
```diff
@@ -17,78 +17,44 @@
...
-  implicit val system: ActorSystem                = ActorSystem("Explorer")
-  implicit val executionContext: ExecutionContext = system.dispatcher
+    //Application Level ExecutionContext not shared with Akka-Http or ActorSystem.
+    implicit val ec: ExecutionContext =
+      ExecutionContext.Implicits.global
 
-  val config: Config = ConfigFactory.load()
+    //Explorer is started Asynchronously. Until the server is started Explorer
+    //Thread is daemon mode so block.
+    Try(Await.result(Explorer(), 30.seconds)) match {
```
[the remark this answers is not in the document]
simerplaha: Done [here](https://github.com/alephium/explorer-backend/pull/255#issuecomment-1149638426) 👍🏼

app/src/main/scala/org/alephium/explorer/Explorer.scala
```diff
@@ -0,0 +1,280 @@
...
+      implicit ec: ExecutionContext,
+      blockFlowClient: BlockFlowClient): Future[Seq[Uri]] =
+    if (directCliqueAccess) {
+      blockFlowClient.fetchSelfClique() flatMap {
+        case Right(selfClique) =>
+          //TODO: What happens when peers are empty?
+          //      Should Application even be allowed to start if there were no peers?
+          val peers = urisFromPeers(selfClique.nodes.toSeq)
+          logger.debug(s"Syncing with clique peers: $peers")
+          Future.successful(peers)
```
simerplaha, line 258: @tdroxler is this TODO right? Should we fail boot-up if peers return empty or is some empty check required here?

tdroxler: yes we should fail, if peers are empty something is very wrong, so better to stop.

tdroxler:
mmmmm I found it bit hard to read the `Explorer` file, with all the `managed` and implicit being created, I'm not sure a new comer would understand how our App start. 
Can we imagine having something that look more like a cooking recipe:
```
    def start () {
      for {
        scheduler        <- createScheduler
        databaseConfig   <- createDatabase
        blockCache       <- createBlockCache
        transactionCache <- createTxCache
        _                <- startSyncServices()
        _                <- startHttpSerer()
      } yield {
        ExplorerState(
          ...
          )
      }
    } 
```
I'm a bit affraid we got lost in what's managed, what are the implicits etc.
HDYT?

simerplaha:
`for` comprehension does look pretty. The problem here that structured the rest of the code was how to easily manage multiple resources that automatically terminate on failure and how easily test it. `FutureUtil.managed` does that but it is hard to use with `for` comprehension. 

- To use `for` comprehension we would need something mutable that collects resources in some concurrent-queue and terminates them. Like Scala's [Using.Manager](https://github.com/scala/scala/blob/v2.13.8/src/library/scala/util/Using.scala#L141) which can be implemented to be thread-safe but it's a lot more code and mutation as you can see.
 - [scala-arm](https://github.com/jsuereth/scala-arm#examples) implements the above for us and works well with `for` comprehension but is not actively maintained. 
 - A custom `Resource` monad around `managed` to work with `for` comprehension would be excessive for cosmetic improvement.
 - There are tons of libraries out there that people built for managing resources. 
 
`FutureUtil.managed` is 18 lines of code and immutable. This was simplest solution I could think of and something that is easy to test.

Alternative solution is not to use `FutureUtil.managed` and have the functions close resources manually within themselves. I'm not sure how this will look like - I have not tried. But this would not be a good idea in general, resource management can get tricky and should be abstracted away. 

I'm sure there are other ways to make this improvement which are probably skipping my mind right now.

tdroxler:
>I'm sure there are other ways to make this improvement which are probably skipping my mind right now.

Thx for the explanation, let's move forward with this then and we'll for sure improve it over the time.

tdroxler (approved):

simerplaha:
Hey @tdroxler, gave this a fair amount of thought. Any improvements I think of are an extension to `managed` that adds `Resource/Managed` type so it can be used in for-comprehension which I’m sure you will not find pleasant because - People will still be lost in all the `implicit`.

I haven’t been able to figure out how to remove `managed` and `AsyncCloseable` with a solution that is not total garbage.

As mentioned on Slack, a solution that uses mutation, manual synchronisation and promises is unnecessarily complex and I’m a bit reluctant to implement to an open-source project.

Maybe we can tell the newcomers to look up loan-pattern and type-classes? I dunno 😕 ... not being able to use type-classes is a big loss. 

Anyhoo, this probably needs a fresh pair of eyes. Can this please be assigned to yourself? So it can be implemented however you think is simpler? I've closed the PR #266 which removes the last two calls to `sys.exit`, built using work done in this PR, will rebase on whatever the new solution is 🚀.

Issue #151, opened 2022-02-23
Cache latest blocks and transactions


polarker:
As latest blocks and transactions are queried more often

simerplaha:
Discussed with Thomas. Setting as high-priority because its a subtask for another high-priority task #200.

tdroxler:
@polarker the latest blocks endpoint is using pagination, the default request from the explorer front end is page 1 and size 10. What about caching only this result as 99% of the time it's the one that will be called ? When a user do a request, if he use the default pagination we use the cache, otherwise we do the db request.

polarker:
There could be two levels of caches here:
1. cache block&transactions based on keys (hash should be enough)
2. cache the list of blocks for the first page, but the list should be expired pretty quick

Maybe we could start with item 1 and leave item 2 for the future. WDYT?

tdroxler:
I'm not sure 1. will bring much for the speed feeling, querying a single block or transaction isn't a bottleneck, but it's anyway good to have it.

point 2 is really to improve the feeling on the home page, as you reported the other day we clearly have to wait for the `Latest Blocks` section when landing on https://explorer.alephium.org/#/ and yes it should expire pretty quick, I guess every 5 seconds as it's our sync period.

polarker:
Item 1 should improve query of latest transaction, as far as I remember transaction query needs join. But I agree that both block and TX query are not bottleneck

That's a nice point for item 2. I already forgot why I raised this issue 😅. Are you going to use the async cache or Caffeine to implement it? Async cache might be a problem if the first page is not queried that frequently.

PR #261, opened 2022-06-10
Feature/async cache block count

### LINKED ISSUE #260: Use `AsyncReloadingCache` for block count
@tdroxler @simerplaha As far as I see, it's not using async cache

tdroxler:
resolves #260

I still need to do a manual test with prod data, but will do this monday.

I took the opportunity to remove some unused code! less code to maintain :muscle:

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
+++ b/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
@@ -17,6 +17,7 @@
 package org.alephium.explorer.cache
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
 import scala.jdk.FutureConverters._
 
 import com.github.benmanes.caffeine.cache.{AsyncCacheLoader, Caffeine}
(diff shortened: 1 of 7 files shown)


simerplaha (approved):

polarker (approved):

app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
```diff
@@ -95,12 +96,9 @@ object BlockCache {
...
 
-    val cacheRowCount: CaffeineAsyncCache[Query[_, _, Seq], Int] =
-      CaffeineAsyncCache.rowCountCache(DBRunner(dc)) {
-        Caffeine
-          .newBuilder()
-          .refreshAfterWrite(5, java.util.concurrent.TimeUnit.SECONDS)
-      }
+    val cacheRowCount: AsyncReloadingCache[Int] = AsyncReloadingCache(0, 10.seconds) { _ =>
+      run(BlockQueries.mainChainQuery.length.result)
+    }
```
simerplaha, line 101: @tdroxler this cache is loading synchronously, so for the first 10.seconds `0` will be returned the clients. We added `AsyncReloadingCache.reloadNow` for this use-case so we get guarantee that the cache is populated before the boot-up is complete. 

But it's not being used so I'm assuming this how you'd want #151 resolved as well? With `Seq.empty[BlockEntryLite]` as cache value for the first n seconds?

tdroxler: mm yeah it would be better to use `reloadNow`. What I did in the `TransationCache` was [to extends it as a Service and invalidate the cache in the startSelfOnce function](https://github.com/alephium/explorer-backend/blob/3d3c9bfb9f8060f0488ca3129d65a59807bcf300/app/src/main/scala/org/alephium/explorer/cache/TransactionCache.scala#L45-L60), maybe we could use `reloadNow` instead? but anyway we could also add `BlockCache extends Service` so we wait that the block cache is populated to start the app.

simerplaha: To avoid the same story again. Probably better to leave this to you to however you like/prefer to use `reloadNow` with `Service`.

PR #262, opened 2022-06-10
Fix config

polarker:
We'd better read configs with a lib


tdroxler:
>We'd better read configs with a lib 

ah yes I wanted to create a ticket but totally forgot, thx for the reminder.

tdroxler (approved):

Issue #196, opened 2022-04-10
Configurable logging


polarker:

killerwhile:
Ideally, log verbosity should be configurable at startup via an environment variable, and at runtime using a REST endpoint (or similar).
The rationale behind changing the verbosity at runtime is to allow debugging a process in a degraded state without requiring a restart of the process, i.e. resetting the state.

polarker:
> The rationale behind changing the verbosity at runtime is to allow debugging a process in a degraded state without requiring a restart of the process, i.e. resetting the state.

Interesting practice! I guess we will need to support this for full node as well.

PR #265, opened 2022-06-14
Add two endpoints to manage log level

tdroxler:
Resolves: #196

`update-global-loglevel` takes a single string to update our global log
level (e.g: DEBUG)

`update-log-config` takes a full `logback.xml` file

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/api/UtilsEndpoints.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/UtilsEndpoints.scala
@@ -28,8 +28,23 @@ trait UtilsEndpoints extends BaseEndpoint with QueryParams {
       .tag("Utils")
       .in("utils")
 
+  private val logLevels    = List("TRACE", "DEBUG", "INFO", "WARN", "ERROR")
+  private val logLevelsStr = logLevels.mkString(", ")
+
   val sanityCheck: BaseEndpoint[Unit, Unit] =
     utilsEndpoint.put
       .in("sanity-check")
       .description("Perform a sanity check")
+
+  val changeGlobalLogLevel: BaseEndpoint[String, Unit] =
+    utilsEndpoint.put
+      .in("update-global-loglevel")
+      .in(plainBody[String].validate(Validator.enumeration(logLevels)))
+      .description(s"Update global log level, accepted: $logLevelsStr")
+
+  val changeLogConfig: BaseEndpoint[String, Unit] =
+    utilsEndpoint.put
+      .in("update-log-config")
+      .in(plainBody[String])
+      .description("Update logging file, only logback.xml is accepted")
 }
--- a/app/src/main/scala/org/alephium/explorer/api/UtilsEndpoints.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/UtilsEndpoints.scala
@@ -17,8 +17,11 @@
 package org.alephium.explorer.api
 
 import sttp.tapir._
+import sttp.tapir.generic.auto._
 
+import org.alephium.api.{alphJsonBody => jsonBody}
 import org.alephium.explorer.api.BaseEndpoint
+import org.alephium.explorer.api.model.LogbackValue
 
 // scalastyle:off magic.number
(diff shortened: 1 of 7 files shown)


app/src/main/scala/org/alephium/explorer/web/UtilsServer.scala
```diff
@@ -39,5 +47,58 @@ class UtilsServer()(implicit ec: ExecutionContext,
...
+        Try(configurator.doConfigure(stream)) match {
+          case Success(_) =>
+            logger.info("Logback configuration updated OK.")
+            Right(())
+          case Failure(e) =>
+            val init = new ContextInitializer(logbackCtx)
+            init.autoConfig
+            logger.warn(
+              s"Cannot apply logback configuation, Logback configuration reset to what is defined on the class path or system properties: ${e.getMessage}")
+            Left(ApiError.BadRequest(s"Cannot apply logback configuation: ${e.getMessage}"))
```
simerplaha, line 96: Being a user facing API we might as well fix typos: `configuration` :) Also at 2 other places in this file.

app/src/main/scala/org/alephium/explorer/api/UtilsEndpoints.scala
```diff
@@ -28,8 +28,23 @@ trait UtilsEndpoints extends BaseEndpoint with QueryParams {
...
+    utilsEndpoint.put
+      .in("update-global-loglevel")
+      .in(plainBody[String].validate(Validator.enumeration(logLevels)))
+      .description(s"Update global log level, accepted: $logLevelsStr")
+
+  val changeLogConfig: BaseEndpoint[String, Unit] =
+    utilsEndpoint.put
+      .in("update-log-config")
+      .in(plainBody[String])
+      .description("Update logging file, only logback.xml is accepted")
```
simerplaha, line 49: `update-global-loglevel` looks nice and validated 👍🏼. Just confirming that `update-log-config` is safe? (access to totally update a config file thats doing disk IO).

simerplaha (commented):
Very cool. 👍🏼

app/src/main/scala/org/alephium/explorer/api/UtilsEndpoints.scala
tdroxler: it only updates the file in runtime, it doesn't change the  disk version. Maybe we should use another endpoint path?

simerplaha: Oh I should've been more clearer. I didn't mean the disk version. Just meant to ask if it's safe to allow total web access to a config file that configures logging IO (something that does disk operations) which is only validated by the logback's parser?

I have no idea. When it comes to security, it favours coding defensively so in the past for similar situations I've had the config file sitting somewhere outside on the same server and load it on boot-up. But maybe that's just me being paranoid. 

Configuring logging via web API is very nice and convenient and if it's safe in terms of security then this looks great 👍🏼 🚀

tdroxler: >config file that configures logging IO (something that does disk operations)

Oh I haven't thought of that. We are restricting the access to those endpoints in our load balancer, but still it might not be a good pratice. Any opinion @killerwhile ?

Maybe we could have a safer version where you can only pass a list of values:
```
[
  {
    "name": "akka.http.impl.engine.client",
    "level":"WARN"
    }
]
```

tdroxler:
in the mean time I also added a `EXPLORER_LOG_LEVEL` env variable just for you @killerwhile :kissing_heart:

tdroxler:
I updated the second endpoint as described, I gave it some manual tests, it works like a charm. Eg:
```
[
  {
    "name": "slick",
    "level": "DEBUG"
  },
 {
    "name": "org.alephium.explorer",
    "level": "ERROR"
  }
]
```

app/src/main/scala/org/alephium/explorer/web/UtilsServer.scala
tdroxler: done

simerplaha (approved):
👍🏼

Issue #259, opened 2022-06-09
Add `close` to `EndpointSender` in `alephium-http` jar dependency


simerplaha:
`EndpointSender` is missing `close` API so there is no way to shutdown `BlockFlowClient`'s sttp client. 

Boot-up failures in PR #255 will not shutdown the JVM because sttp keeps the `Thread` alive.

I've added a commit to this issue to show calling [close in the client](https://github.com/alephium/explorer-backend/commit/240f586f753a1ae769456fcd765dd0de1aa4803b#diff-09e969ecf11f3aa10e8c12c2bf2c91b64ddcc2be147652097cb2e7e4b73b15d4R100) via private method invocation/reflection.

Or is there another way to close the client?

tdroxler:
done in 1.4.1

PR #267, opened 2022-06-15
#259: Added reflection based `close` to `BlockFlowClient`

simerplaha:
- Temporarily solution for #259
- Added reflection based `close` to `BlockFlowClient`


tdroxler:
Thx a lot, let's keep it in our pocket in case we need to deploy explorer-backend before https://github.com/alephium/alephium/pull/648 is merged. No need to merge now.

tdroxler:
Commit is included in https://github.com/alephium/explorer-backend/pull/272

Issue #243, opened 2022-05-23
Kill explorer-backend if cannot access the node


tdroxler:
Currently if the explorer-backend cannot reach the node while syncing,  we just try to call it until it's up again.

We cannot assume the new IP of the node will be the same, especially now that we will move to Kubernetes. The safer is just to kill the explorer-backend and let it restart it.

PR #275, opened 2022-06-21
Stop explorer if cannot acces the node

tdroxler:
An attempt to tackle #243 and #246


app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -117,76 +107,74 @@ object BlockFlowClient {
...
         .send(endpoint, a, uri"${uri.toString}")
-        .map(_.left.map(_.detail))
+        .flatMap {
+          case Right(res) => Future.successful(res)
+          case Left(error) =>
+            logger.error(error.detail)
+            Future.failed(NodeApiError(error.detail))
+        }
+        .recoverWith { error =>
+          Future.failed(UnreachableNode(error.getMessage))
```
simerplaha, line 120: Including the exception message (`error.getMessage`) should we also keep the stacktrace of `error`? So we get failed `send` request's call-stack in the logs?

`UnreachableNode(error)`

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
```diff
@@ -137,14 +137,10 @@ case object BlockFlowSyncService extends StrictLogging {
...
-      case Right(multiChain) =>
-        Future
-          .sequence(multiChain.map(insertBlocks))
-          .map(_.sum)
-      case Left(error) =>
-        logger.error(error)
-        Future.successful(0)
+    blockFlowClient.fetchBlocks(from, to, uri).flatMap { multiChain =>
+      Future
+        .sequence(multiChain.map(insertBlocks))
```
simerplaha, line 142: Should `insertBlocks` here be transactional? Confirming for #206.

simerplaha (commented):
👍🏼

simerplaha:
Just checking: `sys.exit` still exists in `Migration` and `BlockFlowService`. By tackle I'm assuming it's not resolved yet? Or we ok with `sys.exit` in those places?

tdroxler:
> Just checking: `sys.exit` still exists in `Migration` and `BlockFlowService`. By tackle I'm assuming it's not resolved yet? Or we ok with `sys.exit` in those places?

yes #246 isn't completed yet.

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
tdroxler: yes I think you right, that could be problematic if something fail in the middle of the insert, when we restart we might miss some blocks. 
While syncing from scratch we usually sync around 500 blocks at a time, do you think it could be an issue to use `transitionally`? when it's up-to-dae, it's around 5 blocks at a time so it's fine.

tdroxler: thx for the explanation! I think it's fine to use `transactionnally`. the large insert happen only during the `sync-from-scratch` which is usually done "offline".

simerplaha: Sounds great 👍🏼

app/src/main/scala/org/alephium/explorer/error/ExplorerError.scala
```diff
@@ -37,22 +37,27 @@ sealed trait FatalSystemExit extends ExplorerError
 object ExplorerError {
 
   /******** Group: [[FatalSystemExit]] ********/
-  final case class FailedToFetchSelfClique(remoteMessage: String)
-      extends Exception(s"Could not fetch self-clique: $remoteMessage.")
+  final case class UnreachableNode(cause: Throwable)
+      extends Exception(s"Could not reach node: $cause.")
```
simerplaha, line 41: I'm not sure how exceptions in Strings will be logged. Probably use the default constructor parameter to pass `cause` in instead of embedding `Throwable` in the String? 

``` scala 
extends Exception("Could not reach node.", cause)
```

Not nitpicking just trying to get this tiny stuff right so it makes debugging easier incase there is a blow up on the server.

tdroxler: much better, it's fixed

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
tdroxler: I linked the discussion in #206

simerplaha: > yes I think you right, that could be problematic if something fail in the middle of the insert, when we restart we might miss some blocks.

Just going based on your comment here. Not having this being a transaction is a bug as it will lead to invalid database state. Since you've marked it as resolved I guess you ok with merging it.

simerplaha (approved):

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
tdroxler: Yes the bug was still there before, so we'll tackle it in a new PR.

tdroxler: I'll probably work on it today

PR #276, opened 2022-06-21
Add docker workflow for Apple M1 architecture

### LINKED ISSUE #273: Update release workflow to support more platform
`sbt docker` does not build images for arches like Apple M1. Let's use the same workflow as `alephium`.

ref: https://github.com/alephium/alephium/blob/master/.github/workflows/docker-buildx.yml

tdroxler:
Resolves  #273

### DIFF
--- a/.github/workflows/docker-buildx.yml
+++ b/.github/workflows/docker-buildx.yml
@@ -1,0 +1,51 @@
+name: Build and publish docker image
+
+on:
+  workflow_run:
+    workflows: ["Release"]
+    types:
+      - completed
+
+jobs:
+  buildx_and_push_to_registry:
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.workflow_run.head_branch }}
+
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+
+      - name: Get the version
+        id: get_version
+        run: |
+          version=$(git describe --tags --abbrev=0)
+          echo $version
+          echo ${version:1}
+          echo ::set-output name=VERSION::$version
+          echo ::set-output name=VERSION-NO-V::${version:1}
+        shell: bash
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and publish docker image
+        if: ${{ !contains(steps.get_version.outputs.VERSION, '-') }}
+        uses: docker/build-push-action@v2
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-buildx.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: ${{ !contains(steps.get_version.outputs.VERSION, '-') }}
+        uses: docker/build-push-action@v2
+        with:
+          context: docker/release
+          file: ./docker/release/Dockerfile.release
+          build-args: RELEASE=${{ steps.get_version.outputs.VERSION-NO-V }}
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          tags: |
+            docker.io/alephium/explorer-backend:${{ steps.get_version.outputs.VERSION }}
+            docker.io/alephium/explorer-backend:latest
```
polarker, line 49: There will be a race condition if this pushes at the same time with `docker.yml`.

Let's change this line in `build.sbt`:
```
-      val baseImageName = "alephium/explorer-backend"
+     val baseImageName = "alephium/dev-explorer-backend"
```

.github/workflows/docker-buildx.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and publish docker image
+        if: ${{ !contains(steps.get_version.outputs.VERSION, '-') }}
+        uses: docker/build-push-action@v2
+        with:
+          context: docker/release
+          file: ./docker/release/Dockerfile.release
```
polarker, line 44: We will need to have the folder `docker/release` as in full node

tdroxler: ooooh I didn't understand this like that. let me check

polarker: It basically takes the jar file and the docker folder to build a new image. So we will need to adapt the docker file a bit.

tdroxler: Done, I did run a `docker build -f Dockerfile.release .` and then `docker run <image_id>` everything seems fine (I cannot to PG, but it's the same with the official docker image)

polarker: I will give it a try tomorrow

docker/release/Dockerfile.release
```diff
@@ -0,0 +1,18 @@
...
+ARG RELEASE=0.0.0
+
+RUN curl -L https://github.com/alephium/explorer-backend/releases/download/v${RELEASE}/explorer-backend-${RELEASE}.jar -o /explorer-backend.jar
+
+COPY entrypoint.sh /entrypoint.sh
+
+# http
+EXPOSE 9090
+
+ENV JAVA_NET_OPTS="-Djava.net.preferIPv4Stack=true"
```
polarker, line 13: Let's add `USER nobody` before this line

polarker:
I tested it locally, it works well. Just one minor comment.

polarker (approved):

Issue #210, opened 2022-04-22
Improve our usage of `lazy val`


tdroxler:
Linked to that comment https://github.com/alephium/explorer-backend/pull/199#pullrequestreview-948027384

PR #279, opened 2022-06-27
#210: Remove `lazy val` 

simerplaha:
- Removed `AppStateSchema` extending `SchemaMainChain` because it does not create `main_chain` column or index. 
- #210: Removed `lazy val` from `create*Index` because they are not concurrently initialised or need to be kept alive throughout the application. Should be discarded by GC after DB is initialised.


tdroxler (approved):
:rocket: thx <3

PR #282, opened 2022-06-29
Optimize address' transactions query

tdroxler:
Introduce some new fields into `inputs` table to avoid any `JOIN` 

This feature will require a new full re-sync, thus I reset the migrations in a commit.

First 5 commits are minor ones.

You can have more details in each commit's comment.

I haven't activated yet the new feature, I'm waiting a full re-sync locally to validate everything is fine. Once it's done I'll remove previous function and replace with the new ones.

here is a small spoiler:
```
DBBenchmark.getInputsFromTxs         thrpt        363.136          ops/s
DBBenchmark.getInputsFromTxsNoJoin   thrpt       1294.185          ops/s
DBBenchmark.getOutputsFromTxs        thrpt        394.912          ops/s
DBBenchmark.getOutputsFromTxsNoJoin  thrpt        964.500          ops/s
```

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
@@ -101,16 +101,16 @@ object InputQueries {
   }
 
   // format: off
-  def inputsFromTxsNoJoin(hashes: Seq[(BlockEntry.Hash,Transaction.Hash)]):
+  def inputsFromTxsNoJoin(hashes: Seq[(Transaction.Hash,BlockEntry.Hash)]):
     DBActionR[Seq[(Transaction.Hash, Int, Int, Hash, Option[String], Option[Address], Option[U256], Option[Seq[Token]])]] = {
   // format: on
     if (hashes.nonEmpty) {
       val values =
-        hashes.map { case (blockHash, txHash) => s"('\\x$blockHash','\\x$txHash')" }.mkString(",")
+        hashes.map { case (txHash, blockHash) => s"('\\x$txHash','\\x$blockHash')" }.mkString(",")
       sql"""
     SELECT tx_hash, input_order, hint, output_ref_key, unlock_script, output_ref_address, output_ref_amount, output_ref_tokens
     FROM inputs
-    WHERE (block_hash, tx_hash) IN (#$values)
+    WHERE (tx_hash, block_hash) IN (#$values)
     """.as
     } else {
       DBIOAction.successful(Seq.empty)
@@ -118,13 +118,13 @@ object InputQueries {
   }
 
   @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
-  def getInputsQuery(blockHash: BlockEntry.Hash, txHash: Transaction.Hash)
+  def getInputsQuery(txHash: Transaction.Hash, blockHash: BlockEntry.Hash)
     : DBActionSR[(Int, Hash, Option[String], Option[Address], Option[U256], Option[Seq[Token]])] = {
     sql"""
     SELECT hint, output_ref_key, unlock_script, output_ref_address, output_ref_amount, output_ref_tokens
     FROM inputs
-    WHERE block_hash = $blockHash
-    AND tx_hash = $txHash
+    WHERE tx_hash = $txHash
+    AND block_hash = $blockHash
     ORDER BY input_order
     """.as
   }
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
@@ -322,13 +322,13 @@ object OutputQueries {
   }
 
   // format: off
-  def outputsFromTxsNoJoin(hashes: Seq[(BlockEntry.Hash,Transaction.Hash)]):
+  def outputsFromTxsNoJoin(hashes: Seq[(Transaction.Hash,BlockEntry.Hash)]):
   DBActionR[Seq[(Transaction.Hash, Int, OutputEntity.OutputType, Int, Hash, U256, Address,
     Option[Seq[Token]],Option[TimeStamp], Option[ByteString], Option[Transaction.Hash])]] = {
   // format: on
     if (hashes.nonEmpty) {
       val values =
-        hashes.map { case (blockHash, txHash) => s"('\\x$blockHash','\\x$txHash')" }.mkString(",")
+        hashes.map { case (txHash, blockHash) => s"('\\x$txHash','\\x$blockHash')" }.mkString(",")
       sql"""
     SELECT
       outputs.tx_hash,
@@ -343,7 +343,7 @@ object OutputQueries {
       outputs.message,
       outputs.spent_finalized
     FROM outputs
-    WHERE (outputs.block_hash, outputs.tx_hash) IN (#$values)
+    WHERE (outputs.tx_hash, outputs.block_hash) IN (#$values)
     """.as
     } else {
       DBIOAction.successful(Seq.empty)
@@ -352,15 +352,15 @@ object OutputQueries {
 
   // format: off
   @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
-  def getOutputsQuery(blockHash: BlockEntry.Hash, txHash: Transaction.Hash):
+  def getOutputsQuery(txHash: Transaction.Hash,blockHash: BlockEntry.Hash):
   DBActionSR[(OutputEntity.OutputType, Int, Hash, U256, Address, Option[Seq[Token]],
     Option[TimeStamp], Option[ByteString], Option[Transaction.Hash])] = {
   // format: on
     sql"""
         SELECT output_type, hint, key, amount, address, tokens, lock_time, message, spent_finalized
         FROM outputs
-        WHERE block_hash = $blockHash
-        AND tx_hash = $txHash
+        WHERE tx_hash = $txHash
+        AND block_hash = $blockHash
         ORDER BY output_order
       """.as
   }
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
@@ -221,7 +221,7 @@ object TransactionQueries extends StrictLogging {
   def getTransactionsSQL(txHashesTs: Seq[(Transaction.Hash, BlockEntry.Hash, TimeStamp, Int)])(
       implicit ec: ExecutionContext): DBActionR[Seq[Transaction]] = {
     if (txHashesTs.nonEmpty) {
-      val hashes   = txHashesTs.map { case (txHash, blockHash, _, _) => (blockHash, txHash) }
+      val hashes   = txHashesTs.map { case (txHash, blockHash, _, _) => (txHash, blockHash) }
       val txHashes = txHashesTs.map(_._1)
       for {
         inputs  <- inputsFromTxsSQL(txHashes)
@@ -238,7 +238,7 @@ object TransactionQueries extends StrictLogging {
   def getTransactionsNoJoin(txHashesTs: Seq[(Transaction.Hash, BlockEntry.Hash, TimeStamp, Int)])(
       implicit ec: ExecutionContext): DBActionR[Seq[Transaction]] = {
     if (txHashesTs.nonEmpty) {
-      val hashes = txHashesTs.map { case (txHash, blockHash, _, _) => (blockHash, txHash) }
+      val hashes = txHashesTs.map { case (txHash, blockHash, _, _) => (txHash, blockHash) }
       for {
         inputs  <- inputsFromTxsNoJoin(hashes)
         outputs <- outputsFromTxsNoJoin(hashes)
@@ -322,15 +322,15 @@ object TransactionQueries extends StrictLogging {
   }
 
   def gasFromTxsSQL(
-      hashes: Seq[(BlockEntry.Hash, Transaction.Hash)]
+      hashes: Seq[(Transaction.Hash, BlockEntry.Hash)]
   ): DBActionR[Seq[(Transaction.Hash, Int, U256)]] = {
     if (hashes.nonEmpty) {
       val values =
-        hashes.map { case (blockHash, txHash) => s"('\\x$blockHash','\\x$txHash')" }.mkString(",")
+        hashes.map { case (txHash, blockHash) => s"('\\x$txHash','\\x$blockHash')" }.mkString(",")
       sql"""
     SELECT hash, gas_amount, gas_price
     FROM transactions
-    WHERE (block_hash, hash) IN (#$values)
+    WHERE (hash, block_hash) IN (#$values)
     """.as
     } else {
       DBIOAction.successful(Seq.empty)
@@ -344,8 +344,8 @@ object TransactionQueries extends StrictLogging {
       gasAmount: Int,
       gasPrice: U256)(implicit ec: ExecutionContext): DBActionR[Transaction] =
     for {
-      ins  <- getInputsQuery(blockHash, txHash)
-      outs <- getOutputsQuery(blockHash, txHash)
+      ins  <- getInputsQuery(txHash, blockHash)
+      outs <- getOutputsQuery(txHash, blockHash)
     } yield {
       Transaction(txHash,
                   blockHash,
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/InputSchema.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/InputSchema.scala
@@ -50,7 +50,7 @@ object InputSchema extends SchemaMainChain[InputEntity]("inputs") {
     def timestampIdx: Index        = index("inputs_timestamp_idx", timestamp)
     def outputRefAddressIdx: Index = index("inputs_output_ref_address_idx", outputRefAddress)
     def inputsBlockHashTxHashIdx: Index =
-      index("inputs_block_hash_tx_hash_idx", (blockHash, txHash))
+      index("inputs_tx_hash_block_hash_idx", (txHash, blockHash))
 
     def * : ProvenShape[InputEntity] =
       (blockHash,
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/TransactionSchema.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/TransactionSchema.scala
@@ -51,8 +51,6 @@ object TransactionSchema extends SchemaMainChain[TransactionEntity]("transaction
     def blockHashIdx: Index = index("txs_block_hash_idx", blockHash)
     def chainFromIdx: Index = index("txs_chain_from_idx", chainFrom)
     def chainToIdx: Index   = index("txs_chain_to_idx", chainTo)
-    def txsBlockHashTxHashIdx: Index =
-      index("txs_block_hash_tx_hash_idx", (blockHash, hash))
 
     def * : ProvenShape[TransactionEntity] =
       (hash,
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
@@ -56,7 +56,7 @@ object OutputSchema extends SchemaMainChain[OutputEntity]("outputs") {
     def addressIdx: Index   = index("outputs_address_idx", address)
     def timestampIdx: Index = index("outputs_timestamp_idx", timestamp)
     def outputsBlockHashTxHashIdx: Index =
-      index("outputs_block_hash_tx_hash_idx", (blockHash, txHash))
+      index("outputs_tx_hash_block_hash_idx", (txHash, blockHash))
 
     def * : ProvenShape[OutputEntity] =
       (blockHash,
--- a/benchmark/src/main/scala/org/alephium/explorer/benchmark/db/state/AddressReadState.scala
+++ b/benchmark/src/main/scala/org/alephium/explorer/benchmark/db/state/AddressReadState.scala
@@ -64,10 +64,10 @@ class AddressReadState(val db: DBExecutor)
   var blocks: Array[BlockEntity] = _
 
   lazy val hashes = Random
-    .shuffle(blocks.flatMap(_.transactions.map(tx => (tx.blockHash, tx.hash))).toMap)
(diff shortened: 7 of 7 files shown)


app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
```diff
@@ -113,6 +113,22 @@ object OutputQueries {
...
+      val values = txHashes.map(hash => s"'\\x$hash'").mkString(",")
+      sql"""
+    SELECT outputs.tx_hash, outputs.output_order, outputs.hint, outputs.key,  outputs.amount, outputs.address, outputs.lock_time, outputs.spent_finalized
+    FROM outputs
+    WHERE outputs.tx_hash IN (#$values) AND outputs.main_chain = true
+    """.as
+    } else {
+      DBIOAction.successful(Seq.empty)
+    }
+  }
```
simerplaha, line 130: You reckon there is any benefit in using prepared statement here and for `updateTransactionPerAddresses` in `InputUpdateQueries`? Something like the following? 

```scala
val placeholder = paramPlaceholder(rows = 1, columns = txHashes.size)

val query =
  s"""
  SELECT outputs.tx_hash, outputs.output_order, outputs.hint, outputs.key,  outputs.amount, outputs.address, outputs.lock_time, outputs.spent_finalized
  FROM outputs
  WHERE outputs.tx_hash IN $placeholder AND outputs.main_chain = true"""

val parameters: SetParameter[Unit] =
  (_: Unit, params: PositionedParameters) =>
    txHashes foreach { output =>
      params >> output
  }

SQLActionBuilder(
  queryParts = query,
  unitPConv  = parameters
).as
```
I can confirm that different queries with different parameter sizes use different prepared statements so chances of the same prepared statement being used again for randomly sized inputs is lowered. Although this seems odd because it probably shouldn't. I will try to see if this can be avoided somehow.  

``` 
//Prepared-statement with 11 hashes
(S_3,2022-06-30 13:47:48.789585+10,{bytea,bytea,bytea,bytea,bytea,bytea,bytea,bytea,bytea,bytea,bytea},
  SELECT outputs.tx_hash, outputs.output_order, outputs.hint, outputs.key,  outputs.amount, outputs.address, outputs.lock_time, outputs.spent_finalized
  FROM outputs
  WHERE outputs.tx_hash IN ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) AND outputs.main_chain = true)

//Prepared-statement with 2 hashes
(S_7,2022-06-30 13:47:49.161239+10,{bytea,bytea},
  SELECT outputs.tx_hash, outputs.output_order, outputs.hint, outputs.key,  outputs.amount, outputs.address, outputs.lock_time, outputs.spent_finalized
  FROM outputs
  WHERE outputs.tx_hash IN ($1, $2) AND outputs.main_chain = true)
```

app/src/main/scala/org/alephium/explorer/persistence/queries/InputUpdateQueries.scala
```diff
@@ -0,0 +1,67 @@
...
+        .map {
+          case (address, txHash, blockHash, timestamp, txOrder, mainChain) =>
+            s"('$address', '\\x$txHash', '\\x$blockHash', '${timestamp.millis}', $txOrder, $mainChain)"
+        }
+        .mkString(",\n")
+      sqlu"""
+      INSERT INTO transaction_per_addresses (address, hash, block_hash, block_timestamp, tx_order, main_chain)
+      VALUES #$values
+      ON CONFLICT (hash, block_hash, address) DO NOTHING
+      """
```
simerplaha, line 62: Similar to above. Should we use prepared statement here as well instead of splicing?

app/src/main/scala/org/alephium/explorer/persistence/schema/InputSchema.scala
```diff
@@ -37,13 +37,19 @@ object InputSchema extends SchemaMainChain[InputEntity]("inputs") {
...
 
     def pk: PrimaryKey = primaryKey("inputs_pk", (outputRefKey, blockHash))
 
-    def blockHashIdx: Index    = index("inputs_block_hash_idx", blockHash)
-    def inputsTxHashIdx: Index = index("inputs_tx_hash_idx", txHash)
-    def outputRefKeyIdx: Index = index("inputs_output_ref_key_idx", outputRefKey)
-    def timestampIdx: Index    = index("inputs_timestamp_idx", timestamp)
+    def blockHashIdx: Index       = index("inputs_block_hash_idx", blockHash)
+    def inputsTxHashIdx: Index    = index("inputs_tx_hash_idx", txHash)
+    def outputRefKeyIdx: Index    = index("inputs_output_ref_key_idx", outputRefKey)
```
simerplaha, line 50: Please correct me if I'm wrong but I don't think `outputRefKeyIdx` is needed because this column is already indexed in composite primary key `pk`.

tdroxler: yes it's an older commit, I created #283 to tackle it, as it will require a migrations to remove the index, maybe let's try to remove more at once in one single PR

app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
tdroxler: ah yes I always forgot we can use this, it's much cleaner. But while I'm writing this, I remember there is a size limit of placeholders no (not sure at all)? what if we have a lot of tx_hashes?

simerplaha: Oh good point. We could add a `QuerySplitter.splitIn(...)` but I guess that would be alarming as well because if we have to run such large `select` queries with multiple 1000's (> `Short.MaxValue`) of parameters within the `IN` clause then we maybe need to adjust the schema a little? Probably needs grouping into a single logical unique id instead of having to use the `IN` clause.

Sorry for a very vague reply. With the whole 'double checking' tasks I'm sorta looking at this kinda stuff (schema related). Will have a better idea as that progresses.

tdroxler: actually the number of tx_hash in the `IN` will be limited to [100](https://github.com/alephium/explorer-backend/blob/bee50e537607fd808c8b7eee8c49fd819bbe229d/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala#L25). But maybe it's safer to come with a safer version? as you said it would be alarming if we have > 1000 entries for a select.

app/src/main/scala/org/alephium/explorer/persistence/queries/InputUpdateQueries.scala
tdroxler: oh definitively for the `INSERT` I'll update.

tdroxler: done

app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
```diff
@@ -155,6 +99,22 @@ object InputQueries {
...
+  // format: off
+  def inputsFromTxsNoJoin(txHashes: Seq[Transaction.Hash]):
+    DBActionR[Seq[(Transaction.Hash, Int, Int, Hash, Option[String], Transaction.Hash, Address, U256)]] = {
+  // format: on
+    if (txHashes.nonEmpty) {
+      val values = txHashes.map(hash => s"'\\x$hash'").mkString(",")
+      sql"""
+    SELECT tx_hash, input_order, hint, output_ref_key, unlock_script, output_ref_tx_hash, output_ref_address, output_ref_amount
+    FROM inputs
+    WHERE tx_hash IN (#$values) AND main_chain = true AND output_ref_tx_hash IS NOT NULL
```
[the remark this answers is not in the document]
tdroxler: we might miss some inputs in this very specific edge case:

we query a tx and its inputs have output refs that are not yet synced and/or updated by the `InputUpdateQueries.updateInputs`, which happen every 5 seconds.
In a sense we already have this behavior, with the `JOIN` way, if the output ref isn't in db, we will miss the input.

app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
```diff
@@ -113,6 +113,22 @@ object OutputQueries {
...
   }
 
+  // format: off
+  def outputsFromTxsNoJoin(txHashes: Seq[Transaction.Hash]):
+  DBActionR[Seq[(Transaction.Hash, Int, Int, Hash, U256, Address, Option[TimeStamp], Option[Transaction.Hash])]] = {
+  // format: on
+    if (txHashes.nonEmpty) {
+      val values = txHashes.map(hash => s"'\\x$hash'").mkString(",")
+      sql"""
+    SELECT outputs.tx_hash, outputs.output_order, outputs.hint, outputs.key,  outputs.amount, outputs.address, outputs.lock_time, outputs.spent_finalized
```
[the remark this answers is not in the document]
tdroxler: exactly

app/src/main/scala/org/alephium/explorer/persistence/queries/InputUpdateQueries.scala
```diff
@@ -0,0 +1,67 @@
...
+    sql"""
+      UPDATE inputs
+      SET
+        output_ref_tx_hash = outputs.tx_hash,
+        output_ref_address = outputs.address,
+        output_ref_amount = outputs.amount
+      FROM outputs
+      WHERE inputs.output_ref_key = outputs.key
+      AND inputs.output_ref_tx_hash IS NULL
+      RETURNING outputs.address, inputs.tx_hash, inputs.block_hash, inputs.block_timestamp, inputs.tx_order, inputs.main_chain
```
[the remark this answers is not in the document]
tdroxler: 1. arg damn fork, mmmmm even if we do 
```
      WHERE inputs.output_ref_key = outputs.key
      AND inputs.output_ref_tx_hash IS NULL
      AND outputs.main_chain = true
```
it might be an issue if suddenly there's a reorg on the output-ref :/  

both `outputs.address` and `outputs.amount` are fine, as it's an immutable data. The problem is the `outputs.tx_hash`.... Thinking how we use it, it's only to create this little link in the UI: 
image
which redirect to the corresponding transaction. Maybe we could have a new endpoint: `/transaction-by-output-ref/<output_ref>` that we call when we click on that link. So we avoid a complicated mechanism to keep updated the the output ref tx hash in the input table.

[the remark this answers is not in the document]
polarker: > Maybe we could have a new endpoint: /transaction-by-output-ref/<output_ref> that we call when we click on that link.

This sounds great to me 👍  We only update `outputRefAddress` and `outputRefAmount`, which are both immune to forks.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -202,6 +211,57 @@ object TransactionQueries extends StrictLogging {
...
   }
 
+  def getTransactionsNoJoin(txHashesTs: Seq[(Transaction.Hash, BlockEntry.Hash, TimeStamp, Int)])(
+      implicit ec: ExecutionContext): DBActionR[Seq[Transaction]] = {
+    if (txHashesTs.nonEmpty) {
+      val txHashes = txHashesTs.map(_._1)
+      for {
+        inputs  <- inputsFromTxsNoJoin(txHashes)
+        outputs <- outputsFromTxsNoJoin(txHashes)
+        gases   <- gasFromTxsSQL(txHashes)
```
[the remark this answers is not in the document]
tdroxler: I don't see anything else yet, the tokens are done in the next PR, I'm not sure field like `scriptExecutionOk` coming from the node are useful?

app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
simerplaha: You are right, would be good to be safer but I guess then this would also apply to all our existing queries (around 8 queries) that use `IN` or `NOT IN` clause. 

Maybe something to work on in an independent issue?

simerplaha (approved):
👍🏼

app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala
tdroxler: yep, I created https://github.com/alephium/explorer-backend/issues/288

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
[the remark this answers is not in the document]
polarker: Yes, `scriptExecutionOk` will be necessary to support smart contract analysis later, but it's fine to add it later I guess.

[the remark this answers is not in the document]
tdroxler: ok, I'll maybe create a second PR after this one just to add few stuff like the `scriptExecutionOk` and the `inputSignatures`, `scriptSignatures` (as a `bytea`), just to have them in DB, I wont add them yet to the api model.

app/src/main/scala/org/alephium/explorer/persistence/queries/InputUpdateQueries.scala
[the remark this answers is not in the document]
tdroxler: I pushed 2 new commits to add the endpoint, I did lot's of manual test with the prod data it works great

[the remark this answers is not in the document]
polarker: If one input is joint with 2 outputs, does Postgresql update the row twice?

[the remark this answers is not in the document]
tdroxler: hum good question, as on the first update `AND inputs.output_ref_address IS NULL` won't be true anymore...I would need to try, but if it's updated twice it's not harmful.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -116,6 +116,26 @@ object TransactionQueries extends StrictLogging {
...
     }
 
+  def getOutputRefTransactionAction(key: Hash)(
+      implicit ec: ExecutionContext): DBActionR[Option[Transaction]] = {
+    sql"""
+    SELECT tx_hash
+    FROM outputs
+    WHERE main_chain = true AND key = $key
+    """.as[Transaction.Hash].flatMap { txHashes =>
+      txHashes.headOption match {
```
polarker, line 126: We'd better introduce a utility function for unique result query, so it will log errors when there are more than one query result

app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
```diff
@@ -101,14 +99,14 @@ object InputQueries {
...
+    DBActionR[Seq[(Transaction.Hash, Int, Int, Hash, Option[String], Address, U256)]] = {
   // format: on
     if (txHashes.nonEmpty) {
       val values = txHashes.map(hash => s"'\\x$hash'").mkString(",")
       sql"""
-    SELECT tx_hash, input_order, hint, output_ref_key, unlock_script, output_ref_tx_hash, output_ref_address, output_ref_amount
+    SELECT tx_hash, input_order, hint, output_ref_key, unlock_script, output_ref_address, output_ref_amount
     FROM inputs
-    WHERE tx_hash IN (#$values) AND main_chain = true AND output_ref_tx_hash IS NOT NULL
+    WHERE tx_hash IN (#$values) AND main_chain = true AND output_ref_address IS NOT NULL
```
polarker, line 109: There are two issues with this query:
1. We will miss some inputs when `output_ref_address` is NULL, this will make the built transaction incomplete. (If I want to get all of the inputs&outputs of a TX, I don't care if the address is NULL).
2. `main_chain = true` can cause troubles too.

My suggestion:
1. use `Option[Address]` for `output_ref_address` so that the built transaction will always be complete
2. query inputs based on `(txHash, blockHash)` instead of `mainc_chain = true`. I think that this approach should be applied to all inputs&outputs queries when `(txHash, blockHash)` is available.

app/src/main/scala/org/alephium/explorer/persistence/queries/InputUpdateQueries.scala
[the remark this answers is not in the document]
polarker: Sure, it's not a big problem

tdroxler:
FYI after this PR I will release `v1.8.0-RC1`.

This + the tokens needs to be well tested so I prefer an RC version, I'll deploy a staging env against the mainnet data so we can validate everything. Not sure there are tokens on the mainnet, so I might need to do the same on testnet.

app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
tdroxler: 1. Fully agree, front-end can just not show the address if it's missing, they could fall back to the output ref
2. I'll go back to this once I'm done merging master.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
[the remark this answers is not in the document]
polarker: Sounds great to me!

app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
tdroxler: 2. Went back: Great idea, thinking out loud but it might even allows us to remove the `main_chain` in the inputs/outputs (I would need to check in the code tho).

tdroxler: 1. Done

moving to 2.

tdroxler: 2. Done

I also took the time to add the 3 fields we were discussing above, just in db for now, no need to add them to the api yet.

tdroxler: we unfortunatelly won't be able to totally remove `main_chain`, we need it for example for the [getBalance query](https://github.com/alephium/explorer-backend/blob/61939e74e00eebed165ab43b3b030ced17f601fc/app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala#L375-L391)

tdroxler:
I'm restarted a full sync, I'll try to work on the sanity checks so it's done when the sync is done

app/src/main/scala/org/alephium/explorer/persistence/schema/InputSchema.scala
```diff
@@ -49,6 +49,8 @@ object InputSchema extends SchemaMainChain[InputEntity]("inputs") {
     def outputRefKeyIdx: Index     = index("inputs_output_ref_key_idx", outputRefKey)
     def timestampIdx: Index        = index("inputs_timestamp_idx", timestamp)
     def outputRefAddressIdx: Index = index("inputs_output_ref_address_idx", outputRefAddress)
+    def inputsBlockHashTxHashIdx: Index =
+      index("inputs_block_hash_tx_hash_idx", (blockHash, txHash))
```
polarker, line 53: `txHash` should come before `blockHash`, and let's make sure that query planner uses `txHash` index first.

tdroxler: I swap them all so it's coherent everywhere. Here is the planners:

```
explorer=# EXPLAIN SELECT tx_hash, input_order, hint, output_ref_key, unlock_script, output_ref_address, output_ref_amount, output_ref_tokens
explorer-# FROM inputs
explorer-# WHERE (tx_hash, block_hash) IN (('arst','art'));
                                          QUERY PLAN
----------------------------------------------------------------------------------------------
 Index Scan using inputs_tx_hash_block_hash_idx on inputs  (cost=0.43..8.45 rows=1 width=232)
   Index Cond: ((tx_hash = '\x61727374'::bytea) AND (block_hash = '\x617274'::bytea))
(2 rows)

explorer=# EXPLAIN SELECT outputs.tx_hash, outputs.output_order, outputs.output_type, outputs.hint, outputs.key, outputs.amount, outputs.address, outputs.tokens, outputs.lock_time, outputs.message, outputs.spent_finalized
explorer-# FROM outputs
explorer-# WHERE (outputs.tx_hash, outputs.block_hash) IN (('arst','art'));
                                           QUERY PLAN
------------------------------------------------------------------------------------------------
 Index Scan using outputs_tx_hash_block_hash_idx on outputs  (cost=0.55..8.57 rows=1 width=217)
   Index Cond: ((tx_hash = '\x61727374'::bytea) AND (block_hash = '\x617274'::bytea))
(2 rows)

explorer=# EXPLAIN SELECT hash, gas_amount, gas_price
explorer-# FROM transactions
explorer-# WHERE (hash, block_hash) IN (('arst','art'));
                                    QUERY PLAN
----------------------------------------------------------------------------------
 Index Scan using txs_hash_idx on transactions  (cost=0.43..8.45 rows=1 width=42)
   Index Cond: (hash = '\x61727374'::bytea)
   Filter: (block_hash = '\x617274'::bytea)
(3 rows)
```

so both inputs and outputs queries are using the index, but not the query for the gas, maybe postgres can optimize an other way.

app/src/main/scala/org/alephium/explorer/persistence/schema/TransactionSchema.scala
```diff
@@ -51,8 +51,6 @@ object TransactionSchema extends SchemaMainChain[TransactionEntity]("transaction
     def blockHashIdx: Index = index("txs_block_hash_idx", blockHash)
     def chainFromIdx: Index = index("txs_chain_from_idx", chainFrom)
     def chainToIdx: Index   = index("txs_chain_to_idx", chainTo)
-    def txsBlockHashTxHashIdx: Index =
-      index("txs_block_hash_tx_hash_idx", (blockHash, hash))
```
tdroxler: it's already the Primary key, so no need to add this index

app/src/main/scala/org/alephium/explorer/persistence/schema/InputSchema.scala
polarker: Perfect 👍

polarker (approved):
LGTM. Very cool work 👍

Issue #283, opened 2022-06-30
Double check indexes


tdroxler:
Let's try to review our indexes, some are not needed. Let's use this issue to list them once we find some.
Maybe [this one as discussed here](https://github.com/alephium/explorer-backend/pull/282#discussion_r910591301)

polarker:
Relevant tasks: https://github.com/alephium/explorer-backend/issues?q=is%3Aissue+is%3Aopen+index

polarker:
Given that Postgresql does not use indexes as expected sometimes, let's double check the usage of indexes in our query as well.

PR #284, opened 2022-07-04
#283: Remove unneeded index `deps_hash_idx`

simerplaha:
#283: Removes `deps_hash_idx` from table `block_deps`, already indexed in primary key index `hash_deps_pk`.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/BlockDepsSchema.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/BlockDepsSchema.scala
@@ -31,7 +31,6 @@ object BlockDepsSchema extends Schema[BlockDepEntity]("block_deps") {
     def depOrder: Rep[Int]         = column[Int]("dep_order")
 
     def pk: PrimaryKey = primaryKey("hash_deps_pk", (hash, dep))
-    def hashIdx: Index = index("deps_hash_idx", hash)
     def depIdx: Index  = index("deps_dep_idx", dep)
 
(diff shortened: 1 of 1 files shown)


app/src/main/scala/org/alephium/explorer/persistence/schema/BlockDepsSchema.scala
```diff
@@ -31,7 +31,6 @@ object BlockDepsSchema extends Schema[BlockDepEntity]("block_deps") {
     def depOrder: Rep[Int]         = column[Int]("dep_order")
 
     def pk: PrimaryKey = primaryKey("hash_deps_pk", (hash, dep))
-    def hashIdx: Index = index("deps_hash_idx", hash)
```
tdroxler, line 34: I always thought that a tuple primary key was indexing the tuple as a whole, then we should be able to remove `deps_dep_idx` too no?

simerplaha: Postgres doesn't seem to use the primary-key index if we query `dep` directly so we would need `deps_dep_idx` for those queries.

tdroxler (approved):

app/src/main/scala/org/alephium/explorer/persistence/schema/BlockDepsSchema.scala
polarker: SQL index is very counter-intuitive for me, it's hard to tell if they are used or not without using query planner.

PR #286, opened 2022-07-04
#206: Transactional `BlockDao.updateInputs`

simerplaha:
- #206 - Updated `BlockDao.updateInputs` to be transactional.
- Added test-cases for above.
- Updated query `insertTxPerAddressFromInput` to use primary key constraint `txs_per_address_pk`.


app/src/main/scala/org/alephium/explorer/persistence/queries/InputQueries.scala
```diff
@@ -115,7 +115,7 @@ object InputQueries {
     sqlu"""
       INSERT INTO transaction_per_addresses (address, hash, block_hash, block_timestamp, tx_order, main_chain)
       (SELECT address, ${input.txHash}, ${input.blockHash}, ${input.timestamp}, ${input.txOrder}, main_chain FROM outputs WHERE key = ${input.outputRefKey})
-      ON CONFLICT (hash, block_hash, address) DO NOTHING
```
tdroxler, line 118: :+1: much better thx

tdroxler (approved):
thx, this will probably be removed with https://github.com/alephium/explorer-backend/pull/282 but I'll make sure to grab those changes with it.

PR #287, opened 2022-07-05
Switch to `AnyWordSpec`

simerplaha:
Discussed this with Thomas a while back but never got around to it.

We've been using `AnyWordSpec` for new test-cases and its getting a bit repetitive to extend `AnyWordSpec` for every new test-case individually. This PR switches `AlephiumSpec` to use `AnyWordSpec`.

`AnyWordSpec` helps writing more descriptive test-cases and grouping multiple cases for a single test which helps a fair bit because our test-cases are growing in number. 

Eg: 
```scala
"myFunction" should {
  "do something" when {
    "empty" in { //CASE 1
      ???
    }

    "nonempty" in { //CASE 2
      ???
    }
  }
}
```

We lose the keyword `it should` which is kinda redundant anyway because it's required for all tests. 

Replaced existing `AnyFlatSpec` style test-cases 
```scala
it should "do something" in {
  ???
}
```

With

```scala
"do something" in {
  ???
}
```

Thank you Regex!

### DIFF
--- a/app/src/test/scala/org/alephium/explorer/AlephiumActorSpec.scala
+++ b/app/src/test/scala/org/alephium/explorer/AlephiumActorSpec.scala
@@ -24,9 +24,9 @@ import org.scalatest.BeforeAndAfterAll
 class AlephiumActorSpec(val name: String) extends AlephiumActorSpecLike
 
 trait AlephiumActorSpecLike
-    extends TestKitBase
+    extends AlephiumSpec
+    with TestKitBase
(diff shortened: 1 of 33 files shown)


tdroxler (approved):
nice, we get best from two worlds: we can have more precise descriptions of our tests and the `AnyFlatSpec` style is cleaner without all those `it should`

simerplaha:
Woohoo!!

Issue #208, opened 2022-04-22
Double check all joins are based on `main_chain = true`


polarker:
Ref: https://github.com/alephium/explorer-backend/pull/199#discussion_r855931556

tdroxler:
We also need to check we don't have the same issue as in #218

simerplaha:
This can be done in parallel with other 'double check' tasks like #206 & #191. Should they be merged into a single task?

- Checks ensure that all queries are optimised. 
- Implement any missing test-cases to ensure things are transactional and implement tests to avoid any future triple checking.

polarker:
@simerplaha yes, please 🙏  I put it here just in case we forgot it.

tdroxler:
I double checked everything, I think we are good to close this issue.

Issue #123, opened 2022-01-26
Use partial index on `block_headers` table where `main_chain = true`


simerplaha:
Parent task: #120

Target table: `block_headers`
 
## Checks to perform after update

- [ ] Ensure that queries to this table are still using `INDEX ONLY SCAN`
- [ ] Benchmarks should produce better or similar results

tdroxler:
@simerplaha is it still relevant?

simerplaha:
Think this might be an old duplicate of one of the double checking tasks - #208 #283 and probably falls under few other double checking tasks. 

We can close this.

PR #290, opened 2022-07-06
Added `SlickTestUtil.explain()`: For testing Postgres's `EXPLAIN ANALYSE` output

simerplaha:
For issues #208, #123, [comment](https://github.com/alephium/explorer-backend/issues/283#issuecomment-1173800115) and other 'double check' tasks we can write test-cases using `explain()` to assert output of `EXPLAIN ANALYSE` for Slick queries. 

For example:
```scala
val explainQuery = BlockQueries.someQuery(params).explain()

run(explainQuery) should include("Some text from `EXPLAIN ANALYSE` output")
```
I've added an example test for a query in `BlockQueriesSpec` to assert the query uses this `main_chain` index

```sql
index block_headers_full_index
    on block_headers (main_chain asc, block_timestamp desc, hash asc, chain_from asc, chain_to asc, height asc);
 ```

We ok writing such test-cases?

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
@@ -106,8 +106,11 @@ object BlockQueries extends StrictLogging {
     * Fetches all main_chain [[org.alephium.explorer.persistence.schema.BlockHeaderSchema.table]] rows
     */
   def listMainChainHeadersWithTxnNumberSQL(
-      pagination: Pagination): DBActionRWT[Vector[BlockEntryLite]] = {
+      pagination: Pagination): DBActionRWT[Vector[BlockEntryLite]] =
+    listMainChainHeadersWithTxnNumberSQLBuilder(pagination)
+      .as[BlockEntryLite](blockEntryListGetResult)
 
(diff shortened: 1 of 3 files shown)


tdroxler (commented):
oh that's an awesome idea :rocket: even just as a tool when writing a query, instead of always copy/pasting in pg.

Should we have both `explain()` and `explainAnalyse()` function? 

It could be useful to just use `explain()` so we don't need to fill the db with some data, or is there a difference in the output between the two depending of the data?

simerplaha:
> oh that's an awesome idea 🚀 even just as a tool when writing a query, instead of always copy/pasting in pg.

🚀 

> Should we have both `explain()` and `explainAnalyse()` function?
> It could be useful to just use `explain()` so we don't need to fill the db with some data, or is there a difference in the output between the two depending of the data?

Yep sure can have both `explain()` and `explainAnalyse()`. It seems `EXPLAIN ANALYSE` returns same info as `EXPLAIN` plus query benchmarks. We might not need to test benchmarks for all queries so having both functions as you suggested will be nice.

tdroxler (approved):
Can't wait to use it, this will make our life so much easier to check our indexes. THXXXXXX <3

PR #291, opened 2022-07-06
Add a super fast endpoint to check if an address is active (at least 1 tx)

tdroxler:
based on [this discussion](https://github.com/alephium/desktop-wallet/issues/136#issuecomment-1176264853)

This is a feature needed for address discovery in the desktop-wallet

here is the curl time in second: 
maim-1657120048

The query is: 
```
SELECT EXISTS (SELECT 1 FROM transaction_per_addresses WHERE address = $address)
```
it's the fastest thing I could find and the one that pop up the more often in all my searches

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
@@ -96,10 +96,12 @@ trait AddressesEndpoints extends BaseEndpoint with QueryParams {
       .out(jsonBody[Seq[Transaction]])
       .description("List address tokens")
 
-  val isAddressActive: BaseEndpoint[Address, Boolean] =
-    addressesEndpoint.get
-      .in(path[Address]("address")(Codecs.addressTapirCodec))
-      .in("active")
-      .out(jsonBody[Boolean])
-      .description("Is the address active (at least 1 tx)")
+  val areAddressesActive: BaseEndpoint[Seq[Address], Seq[Boolean]] =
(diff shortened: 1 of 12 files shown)


simerplaha (approved):
Neat!!! 🚀 🚀

tdroxler:
I updated the PR with what was discussed in https://github.com/alephium/desktop-wallet/issues/136 , basically it now takes a list of addresses instead of just one

tdroxler:
New benchmark, 1000 addresses take around 0.1s, 4000 address 0.5s.

Which makes me think we should limit the number of addresses.... I'll add this

app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
```diff
@@ -28,6 +28,8 @@ import org.alephium.protocol.Hash
 // scalastyle:off magic.number
 trait AddressesEndpoints extends BaseEndpoint with QueryParams {
 
+  private val activeAddressesMaxSize: Int = 200
```
polarker, line 31: Nit: maybe 4 * 20 since the gas limit of address discovery is 20 by default?

polarker (approved):

app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
tdroxler: discussed this with front-end team, we will go with what you propose to follow the more closely bip44

Issue #223, opened 2022-05-06
Sanity check app before production deployment


polarker:
We might implement this as a Scala app like `tools` in the full node repo.

Here are some of the checks:
* balance query for all addresses
* compare the balance against full node
* check the latency of cached queries
* check main_chain

polarker:
I suggest using Scala instead of shell script, as the check app might get very complicated for the long term

polarker:
More high priority checks:
* block query returns the same list of transactions as full node
* tx query returns the same list of inputs/outputs/meta data as full node (no input/output missing, no duplicated input/output)

polarker:
@tdroxler @simerplaha I changed this to high priority as I feel that complete sanity check is critical for our latest breaking changes.

polarker:
One more nice check following @simerplaha 's great analysis: https://github.com/alephium/explorer-backend/pull/292#issuecomment-1186783387

We could query for the plan outputs of all our queries after the backend is fully synced, and check if indexes are used as expected.

PR #292, opened 2022-07-07
Test: `mainChainQuery` should use `block_headers_main_chain_idx`

simerplaha:
- Added a test (a bit flaky, reason below) to check `mainChainQuery` uses `block_headers_main_chain_idx`.
- Added tag `IndexCheck` incase we want to exclude these test-cases from every build because they can be slow. 
- Added `explain` & `explainAnalyse` for typed queries. 

Postgres randomly (very rarely) chooses not to use `block_headers_main_chain_idx` and prefers sequential scan.

## LOG

Multiple attempts at writing more data after initial failed attempt at using the index does not change Postgres's decision. 

This could possibly be due to the difference in data generated, but I haven't narrowed it down yet. Even though the ratio of `mainChain` vs `!mainChain` is similar and the failure to use the index mostly occurs when the size of `mainChain = true` is slightly larger than `!mainChain`/`mainChain = false`.

I've tried dropping other indexes to see if this behaviour changes but no luck there. 

Here are the logs of the attempts after first fail to use the index in the query. 

> DataSize: 20000 //Postgres uses the index for first iteration
!mainChain = 10146
mainChain = 9854
Bitmap Heap Scan on block_headers  (cost=100.86..1628.46 rows=18760 width=260)
  Recheck Cond: main_chain
  ->  Bitmap Index Scan on block_headers_main_chain_idx  (cost=0.00..96.17 rows=18760 width=0)

> DataSize: 20000 //Next iteration with new data Postgres chooses not to use the index
!mainChain = 9942
mainChain = 10058
Seq Scan on block_headers  (cost=0.00..1652.67 rows=3606 width=235)
  Filter: main_chain

> Re-run explain //Ran Explain with same data - Still the same: Index not used
Seq Scan on block_headers  (cost=0.00..1652.67 rows=3606 width=235)
  Filter: main_chain

> Add more data explain: 40000 //Wrote 20,000 more records. Still no change 
Seq Scan on block_headers  (cost=0.00..2104.25 rows=4591 width=235)
  Filter: main_chain

> Add more data explain: 60000 //Wrote 20,000 more records. Still no change
Seq Scan on block_headers  (cost=0.00..2804.63 rows=6120 width=235)
  Filter: main_chain

> Add more data explain: 80000 //Wrote 20,000 more records. Still no change
Seq Scan on block_headers  (cost=0.00..3505.00 rows=7648 width=235)
  Filter: main_chain

## Benchmarks

To remind ourselves of performance difference: Reran our old benchmark for, with and without `main_chain_index` and with connection pool disabled.

image

Maybe we should run `EXPLAIN` on boot-up or randomly (`SanityCheck`?) to guarantee that the index is being used?

Long shot: An Admin UI would be nice for checking stuff like this.


tdroxler:
I don't know if it helps, but we can easily assume we will always have much much more `main_chain = true`

polarker:
> I don't know if it helps, but we can easily assume we will always have much much more `main_chain = true`

We could assume the percentage of `main_chain = true` is > 99%.

tdroxler:
> We could assume the percentage of `main_chain = true` is > 99%.

From what I read a bit online, it seems that when the query will anyway read most of the data, it prefers to use `Seq Scan` rather than `Index Scan`. Maybe our `main_chain_index` isn't that useful as we just said that it covers 99% of the data.

app/src/test/scala/org/alephium/explorer/persistence/queries/BlockQueriesSpec.scala
```diff
@@ -155,4 +158,42 @@ class BlockQueriesSpec
...
+      run(BlockHeaderSchema.table ++= headers).futureValue
+
+      //build explain query for mainChainQuery
+      run(BlockQueries.mainChainQuery.result.explainFlatten()).futureValue
+    }
+
+    "use block_headers_main_chain_idx" when {
+      "data size is large" taggedAs IndexCheck in {
+        /*
+         * This test is flaky. Postgres randomly choose not to use `block_headers_main_chain_idx`.
```
tdroxler: hum... I'm not fan of adding a flaky test, even tho I understand the all context, this will be too boring to have PR failing because of this. Let's try to see if we really need this `main_chain` in the end.

simerplaha: The test seems necessary, even if it's flaky because without this test we wouldn't have known that this problem exists.

Nice. if we can remove the need for `main_chain` all together that'd be great. But are you referring to `main_chain` column or `main_chain` index?

polarker: Could you try to generate a random dataset with 200 data records and 198 of them are  `main_chain = true`? Let's see if it will flaky still.

simerplaha: Low randomness in data skips the index.

Your test:

| main_chain | count(*) |
|------------|-------|
| false      | 2     |
| true       | 198   |

> Seq Scan on block_headers  (cost=0.00..12.80 rows=140 width=260)
  Filter: main_chain

polarker: Thanks. Could you also paste the two different query analysis results so I could see the exact plans?

simerplaha: //EXPLAIN (Under the table above)
Seq Scan on block_headers  (cost=0.00..12.80 rows=140 width=260)
  Filter: main_chain

//EXPLAIN ANALYZE
Seq Scan on block_headers  (cost=0.00..12.80 rows=140 width=260) (actual time=0.018..0.091 rows=198 loops=1)
  Filter: main_chain
  Rows Removed by Filter: 2
Planning Time: 0.796 ms
Execution Time: 0.139 ms

polarker: Maybe I missed something, but both are using the `main_chain` index right?

simerplaha: They both are using `main_chain` column but not the index because of less variation in data.

The query being analysed is [BlockQueries.mainChainQuery](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala#L46).

We get the same plan output when the query is run directly in Postgres. 

```sql
EXPLAIN ANALYSE
SELECT *
FROM block_headers
WHERE main_chain = true;
```

polarker: Could you show me the different execution plans (EXPLAIN query)? I want to see the difference and might get some new ideas from the plan outputs.

polarker: > We get the same plan output when the query is run directly in Postgres.

@simerplaha please paste the different plan outputs in the flaky test (when test again 200 data records with 198 of them have `main_chain = true`). I cannot guess what are the different outputs, and I don't have the time to get them by myself. Thank you in advance!

polarker:
For the other queries that use `mainChainQuery`, the execution plan might be more unpredictable?

I am thinking that we might create an endpoint to return all of the plan outputs of all of our queries. WDYT?

app/src/test/scala/org/alephium/explorer/persistence/queries/BlockQueriesSpec.scala
simerplaha: Hey Cheng, yep definitely.

Just wanted to run some additional tests supplementing your request to figure out why this index is misbehaving.

Will surely sort this out. Sorry for the delay.

simerplaha:
> For the other queries that use `mainChainQuery`, the execution plan might be more unpredictable?

You are right. We will be 100% sure about this after running those tests. 

> I am thinking that we might create an endpoint to return all of the plan outputs of all of our queries. WDYT?

This is an awesome idea 👍🏼. We will know the importance of this after running those tests. We should have this all figured out in the coming week :)

simerplaha:
A report on trying to figure out the tipping point when Postgres decides not use the `main_chain_idx` index. 

All the following graphs include your suggested test of 198-2 ratio.

## Graph 
- `x-axis` = Total number of rows with `main_chain = true`
- `y-axis` = Total number of rows with `main_chain = false`
- Red plot = `main_chain_idx` not used
- Green plot = `main_chain_idx` used

# Test 1

First iteration starting count of  `main_chain = false` = MAX `main_chain = true` = 0.

AND

Last iteration the final count of  `main_chain = false` = 0 `main_chain = true` = MAX.

## Sample iteration data log when max data size is 200
Iteration0 - Ratio (true, false): 0 : 200\
Iteration1 - Ratio (true, false): 1 : 199\
Iteration2 - Ratio (true, false): 2 : 198\
Iteration3 - Ratio (true, false): 3 : 197\
Iteration4 - Ratio (true, false): 4 : 196\
.\
.\
Iteration196 - Ratio (true, false): 196 : 4\
Iteration197 - Ratio (true, false): 197 : 3\
Iteration198 - Ratio (true, false): 198 : 2\
Iteration199 - Ratio (true, false): 199 : 1\
Iteration200 - Ratio (true, false): 200 : 0

## datasize = 200

image

## datasize = 500

image

## datasize = 1000

image

## datasize = 2000

image

# Test 2

First iteration starting count of  `main_chain = false` = 0 `main_chain = true` = MAX.

AND

Last iteration the final count of  `main_chain = false` = MAX `main_chain = true` = 0.

## datasize = 200

image

## datasize = 500

image

## datasize = 1000

image

## datasize = 2000

image

# Notes

The probability of the index being is used is higher when the data-size is large. But even then, it randomly fails to use the index as shown by red points on the graphs, specially the last graph. 

The behaviour is inconsistent. `Test 1` behaves differently to `Test 2` even when the data is similar. The only difference is the order in which the data is written.

app/src/test/scala/org/alephium/explorer/persistence/queries/BlockQueriesSpec.scala
simerplaha: Does [this](https://github.com/alephium/explorer-backend/pull/292#issuecomment-1186783387) report match what you are asking for? The graphs should give a sense of probability.

polarker:
@simerplaha Thank you for the great analysis 👍  The result is a bit mind-blowing for me honestly.

I don't have any idea to make this test non-flaky. The easier way is probably to check indexes in the "sanity check" task with production data: https://github.com/alephium/explorer-backend/issues/223#issuecomment-1186900475

simerplaha:
## Update 

Moved index-check related code out of test-cases so it could be used as a service. 

Added an endpoint (http://localhost:9090/utils/index-check) to show health of queries. 

`passed: boolean` indicates if the test passed or failed. 

Is the route we want to take? 

image

tdroxler:
> ## Update (very initial version)
> 
> Moved index-check related code out of test-cases so it could be used as a service `IndexChecker` with production data.
> 
> Added an endpoint (http://localhost:9090/utils/index-check) to show health of queries.
> 
> `passed: boolean` indicates if the test passed or failed.
> 
> We ok with this?

I haven't fully checked the code yet, but the output result looks super nice.

simerplaha:
Builds failing due to these index-check flaky tests
- [TransactionQueriesSpec](https://github.com/alephium/explorer-backend/runs/7497558984?check_suite_focus=true#step:5:303)
- [InputQueriesSpec](https://github.com/alephium/explorer-backend/runs/7497172580?check_suite_focus=true#step:5:592)

These will be fixed depending on the outcome of this PR.

polarker:
Hey all, feel free to merge if it looks great. I only gave feedbacks when I have some time.

tdroxler (approved):

PR #294, opened 2022-07-11
Updates to `InputQueries`, `OutputQueries` & `TransactionQueries`

simerplaha:
- Removed splicing from some queries so it gets used as prepared statement. Ignoring to-be deprecated queries like `inputsFromTxsSQL`. 
- Removed large tuples from query output and replaced with types. 
- #283: Removed unneeded indexes from `InputSchema`, `OutputSchema` & `TransactionSchema`.
- Formatted few SQL queries using standard SQL formatter. IntelliJ makes this easier, for folks not using IntelliJ they can use [this](https://www.dpriver.com/pp/sqlformat.htm)?

And other stuff, see commit messages.


app/src/test/scala/org/alephium/explorer/persistence/queries/InputQueriesSpec.scala
```diff
@@ -198,4 +198,28 @@ class InputQueriesSpec
...
+            val query =
+              sql"""
+                   |SELECT *
+                   |FROM inputs
+                   |where output_ref_key = ${input.outputRefKey}
+                   |""".stripMargin
+
+            val explain = run(query.explain()).futureValue.mkString("\n")
+
+            explain should include("Index Scan using inputs_pk on inputs")
```
tdroxler: all those tests are so great :rocket:

tdroxler (approved):
Nice cleaning :rocket: thx a lot

app/src/test/scala/org/alephium/explorer/persistence/queries/InputQueriesSpec.scala
simerplaha: I guess this test case is flaky as well. Build on master is [failing](https://github.com/alephium/explorer-backend/runs/7497172580?check_suite_focus=true#step:5:592). 

Will make changes according to the outcome #292.

PR #297, opened 2022-07-19
Switch `areAddressesActive` to use `UNION`

simerplaha:
# Overview 
Current `areAddressesActive` queries are dispatched concurrently. Cost of "threads used" is relative to the size of input (`Seq[Address]`). 

This PR switches it to use `union` so there is no concurrent IO. 

Actual performance improvement can be seen with `DisabledCP` i.e. nearly 30x.

With `HikariCP` Performance improves over 5x.

# Benchmark
Benchmarks with `DisabledCP` & `HikariCP`

- Top two benchmarks use existing query - `areAddressesActiveAction`
- Last two benchmarks use `union` query - `areAddressesActiveActionUnion`
-  Benchmark data 
    - 400,000 `transaction_per_addresses` rows
    - `40` addresses per query
    - Run time: 1 minute per benchmark

image

# TODO

The query uses `Seq[Address]` as input, this should be a `Set`.

```scala 
def areAddressesActiveAction(addresses: Seq[Address]) = ...
```


tdroxler (approved):
:rocket:

PR #306, opened 2022-07-26
Use `ficus` to manage config

tdroxler:
all the nice `validate*` functions from `ExplorerConfig` were kept and used to create the corresponding `ficus.ValueReader`


simerplaha (approved):
Very cool. Removes a bit of code 👍🏼 . Kinda odd though that with Ficus we have the need to use `.get`. 

LGTM  🚀

tdroxler:
> Very cool. Removes a bit of code 👍🏼 . Kinda odd though that with Ficus we have the need to use `.get`.
> 
> LGTM rocket

yeah Ficus is unfortunately throwing errors, but I find it ok for this usage, as the config is the very first thing that is loaded and if there's an error anyway we can't start the app. 
But that would have been nice from them to have a constructor for `ValueReader` accepting a `Try`

Issue #191, opened 2022-04-01
Use partial index properly


polarker:
* an index for `main_chain` is a huge B-tree with 2 keys, this sounds weird
* alternatively, we create partial indexes for all relevant tables, to save storage size and improve performance (probably slightly)

@simerplaha do think we need to do a benchmark for this task? I am convinced intuitively :)

simerplaha:
Here is the summary of new benchmarks. Hopefully helps confirm which way to proceed.

The outcome is still the same as other benchmarks in other PRs. From the many iterations I've executed (changing order of execution etc ...) I'm unable to definitively confirm which one is faster between full VS partial index. We already know we [save disk space](https://github.com/alephium/explorer-backend/issues/158#issuecomment-1077149490) with partial index.

# Benchmark types
- *NoIndex - With no index
- *FullIndex - Default basic index
- *Reset_10 - Eg: For `main_chain` this will create 1 row with `main_chain = true` for every 10 rows
- *Reset_100 - Eg: For `main_chain` this will create 1 row with `main_chain = true` for every 100 rows
- *Reset_1000 - Eg: For `main_chain` this will create 1 row with `main_chain = true` for every 1000 rows

# main_chain (boolean)

- Full index - `create index full_main_chain_idx on transactions (main_chain);`
- Partial index - `create index partial_main_chain_idx on transactions (main_chain) where main_chain = true;`
- Query benchmarked: `select count(*) from transactions where main_chain = true`

![image](https://user-images.githubusercontent.com/1773953/161744592-3cdc78f7-19a2-4ac6-9af7-e006a1c53727.png)

<details><summary>More benchmarks</summary>
<p>

![image](https://user-images.githubusercontent.com/1773953/161744670-3a82bdf6-6d8d-4d7a-b48c-6a702810a2c9.png)

![image](https://user-images.githubusercontent.com/1773953/161744685-812e93da-611f-4f37-84e9-c383f1e045dc.png)

</p>

</details>

# interval (integer)

- Full index - `create index full_interval_idx on transactions (interval);`
- Partial index - `create index partial_interval_idx on transactions (interval) where interval = 0;`
- Query benchmarked: `select count(*) from transactions where interval = 0`
- `*Reset` benchmarks set `interval` integer value incrementally and set to 0 for every new interval checkpoint. 

![image](https://user-images.githubusercontent.com/1773953/161745085-6b9e7745-b55c-49b4-82fa-61e7a5931fc0.png)

[...]

PR #309, opened 2022-08-01
Added `explain` for query `getTxnHashSQL`

simerplaha:
## Overview
- For #283, #191
- Added check for query `getTxnHash` to `IndexChecker`
- Added field `message` to `ExplainResult` for showing more info. For example: The following image shows the output for all indexes used by the query. 

## JSON
- First is for an `output` with oldest `timestamp`.
- First is for an `output` with latest `timestamp`. 
- Real data used by running sync partially. `173651` rows in `outputs` table, all with `main_chain = true`.

image

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
@@ -52,7 +52,7 @@ object BlockQueries extends StrictLogging {
         queryName  = "mainChainQuery",
         queryInput = "Unit",
         explain    = explain,
-        message    = None,
+        message    = Iterable.empty,
         passed     = explain.mkString contains "block_headers_main_chain_idx"
       )
     }
@@ -129,7 +129,7 @@ object BlockQueries extends StrictLogging {
         queryName  = "listMainChainHeadersWithTxnNumber",
         queryInput = pagination.toString,
         explain    = explain,
-        message    = None,
+        message    = Iterable.empty,
         passed     = explain.mkString contains "block_headers_full_index"
       )
     }
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
@@ -52,6 +52,7 @@ object BlockQueries extends StrictLogging {
         queryName  = "mainChainQuery",
         queryInput = "Unit",
         explain    = explain,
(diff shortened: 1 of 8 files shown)


app/src/main/scala/org/alephium/explorer/persistence/queries/ExplainResult.scala
```diff
@@ -25,4 +25,5 @@ object ExplainResult {
 final case class ExplainResult(queryName: String,
                                queryInput: String,
                                explain: Vector[String],
+                               message: Iterable[String],
```
tdroxler, line 28: a bit counter-intuitive to have an `Iterable` while using `None` as value. I would prefer using an `Option` here, also for the JSON serialization, as the `Iterable` will be a `[]` while `Option` is `null`. We want the second.

app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
```diff
@@ -52,7 +52,7 @@ object BlockQueries extends StrictLogging {
         queryName  = "mainChainQuery",
         queryInput = "Unit",
         explain    = explain,
-        message    = None,
+        message    = Iterable.empty,
```
tdroxler, line 55: oh I see you change here, but still can we have multiple messages? if yes we could renamed the field to `messages`, otherwise we should use `Option`

app/src/main/scala/org/alephium/explorer/persistence/queries/ExplainResult.scala
simerplaha: Switched `Option[String]`.

The reason to use a `Iterable[String]` was so that the output json is more readable. Just like how `explain` being a `Vector[String]` renders the text in JSON just like how Postgres renders it in console. 

## Currently `Iterable[String]` gets this which is more readable 
image

### Using `Option[String]` get this
image

### Using `Option[String]` without newline get this
image

tdroxler: ooooooo my bad, this is indeed way more readable to use the `Iterable` + it makes sense as each item is another message. Then maybe we could go back to `Iterable` but rename to `messages`? HDYT? 
sorry for this back and forth

simerplaha: No worries at all :) Will do. Yep `messages` is clearer will make this change.

tdroxler (approved):
:rocket:

PR #310, opened 2022-08-02
Add missing api model serialization test

tdroxler:
Resolves: #98


simerplaha (approved):
🚀 

Do you prefer splitting test generators into multiple files? Maybe your `Gen[*]` types could be moved to `GenModel`? Or maybe `GenModel` should be removed and merged into this single `Generators` file?

I personally prefer splitting them up. This file is going to grow quick. As it grows IDEs are going to struggle with autocomplete which increases compile time and a pain. 

However you prefer. Your call 📲

Either way it looks good. LGTM.

tdroxler:
yes I can split, maybe in a new `ApiModelGen` as `GenModel` is having the db model, I might rename that second too.

tdroxler:
what's an IDE btw? 


:trollface:

simerplaha:
Hahah You legend.

PR #312, opened 2022-08-03
Feature/improve laziness

tdroxler:
Resolves: https://github.com/alephium/explorer-backend/issues/210

I need your careful expertise for this one @simerplaha :)


simerplaha (approved):
Looks awesome! 🚀 🚀 Yay!

Just added tiny tiny commit to use `flatMap` instead of `.get` (No biggie :))

PR #314, opened 2022-08-04
Fix/issue 142

tdroxler:
Basically we were not using the correct primary key and when we were
inserting multiple uoutput with the same address but a different lock
time, we were overriding the previous one.


simerplaha (approved):
🚀

PR #318, opened 2022-08-09
Cache blocks pages

tdroxler:
Cache blocks based on pagination

The goal is to improve the home page of our explorer.

We trust here the eviction algorithm of Caffeine:

[tinyLFU](https://dl.acm.org/doi/pdf/10.1145/3149371)

As the first page of the explorer is the one that will be called almost
all the time compared to any other pages, it should almost never be
evicted by the cache.

Resolves: #151


tdroxler:
Closing this for now, as you mentioned Simer there are a lot of scenario where it could go wrong and currently the home page doesn't suffer since we have our new instance. I might go back on this latter.

Issue #95, opened 2021-08-27
Endpoint to get paginated mempool txs


polarker:

nop33:
This feature would improve our frontend apps a lot. At the moment, we have no way to get informed about unconfirmed transactions. This means that if the user sends a transaction using the desktop wallet and then closes the app and opens it again, they have no way of knowing the status of their pending transactions.

mvaivre:
Hey guys! Could we start looking at this? It would be very valuable to display pending txs in the explorer and our wallets.

tdroxler:
:+1: I'll start working on that, we already have the pending tx in db so it shouldn't be hard.

nop33:
Reopening due to https://github.com/alephium/explorer-backend/pull/320#issuecomment-1218116053

tdroxler:
Reclosing in favor of https://github.com/alephium/explorer-backend/issues/326
This issue was about a paginated endpoint, for example to have an equivalent of https://cn.etherscan.com/txsPending

PR #320, opened 2022-08-11
Add `/unconfirmed-transactions` endpoint

tdroxler:
We also added a `lastSeen` field to `UnconfirmedTransaction`, the same
way [https://cn.etherscan.com/txsPending][Etherscan] is doing.
This help to sort the txs, but also to see if some txs are in the
mempool for a long time.

Resolves: #95


tdroxler:
@nop33 this new endpoints is just about listing the unconfirmed-transactions, but maybe it's not enough for you to be able to recover the pending tx for a given address. Maybe we should also add an `/addresses/<address>/unconfirmed-transactions` endpoint?

nop33:
> listing the unconfirmed-transactions

like, ALL unconfirmed txs of the mempool? Yes, we'd def need an endpoint to fetch unconfirmed txs of a specific address. Thanks! 🙏

app/src/main/scala/org/alephium/explorer/persistence/dao/UnconfirmedTxDao.scala
```diff
@@ -70,7 +74,36 @@ object UnconfirmedTxDao extends UnconfirmedTxDao {
...
+  }
+
+  def list(pagination: Pagination)(
+      implicit executionContext: ExecutionContext,
+      databaseConfig: DatabaseConfig[PostgresProfile]): Future[Seq[UnconfirmedTransaction]] = {
+    val offset = pagination.offset.toLong
+    val limit  = pagination.limit.toLong
+    val toDrop = offset * limit
+    run(for {
+      txs <- UnconfirmedTxSchema.table.sortBy(_.lastSeen.desc).drop(toDrop).take(limit).result
```
simerplaha, line 91: `lastSeen` might need an index in the table `UnconfirmedTxSchema`?

tdroxler:
> > listing the unconfirmed-transactions
> 
> like, ALL unconfirmed txs of the mempool? Yes, we'd def need an endpoint to fetch unconfirmed txs of a specific address. Thanks! pray

@nop33 the address is only present in the outputs, for performance reasons we won't have the address in the inputs. Which means when you will fetch the endpoint for a given address, you will also receive the tx going TO that address. Is that an issue?

nop33:
> @nop33 the address is only present in the outputs, for performance reasons we won't have the address in the inputs. Which means when you will fetch the endpoint for a given address, you will also receive the tx going TO that address. Is that an issue?

@tdroxler will I be able to differentiate which TX is a "sending pending tx" and which is a "receiving pending tx"? If so, then it's fine!

tdroxler:
> > @nop33 the address is only present in the outputs, for performance reasons we won't have the address in the inputs. Which means when you will fetch the endpoint for a given address, you will also receive the tx going TO that address. Is that an issue?
> 
> @tdroxler will I be able to differentiate which TX is a "sending pending tx" and which is a "receiving pending tx"? If so, then it's fine!

@nop33 from an input we cannot fully guarantee to find back the address, as we have different kind of inputs, depending if it comes from a single address, from a multisig or a script.
But in our wallet case, in a "sending tx", we should be able to find back the address (but I still need to do some test). 

What we could maybe do is in our `UInput` model is to have an optional field `address`, that you could use to define if it's a FROM or TO transaction. If no address is present, it will either be a multisig tx or a script tx, we could then think how to show those kind of txs.

I will experiment a bit and come back to you

simerplaha (approved):
🚀

tdroxler:
I'll merge this PR and create a new issue for the unconfirmed txs per address endpoint

PR #322, opened 2022-08-16
Resolves #298: Reuse ScalaCheck test-data Generators

### LINKED ISSUE #298: Reuse ScalaCheck test-data Generators
`benchmark` module is building its own `DataGenerator` functions. This is duplicating work. 

We should reuse generators written for ScalaCheck from the `app` module.

simerplaha:
- Resolves #298
- Allows reusing ScalaCheck generators from `app` module in benchmarks


tdroxler:
Don't you have to remove [DataGenerator](https://github.com/alephium/explorer-backend/blob/bb9dc6ba10105bd6e74dc6a91aae9eaf35d8d203/benchmark/src/main/scala/org/alephium/explorer/benchmark/db/DataGenerator.scala) to fully complete the task?

simerplaha:
Hmm. Didn't think so. We are not running those benchmarks at all. Documented once in an issue somewhere. Seemed like unnecessary work to port code that's not being used or breaking anything. 

But will do. Since you asked for it :)

tdroxler (approved):

Issue #148, opened 2022-02-19
Use IndexedSeq or ArraySeq or AVector instead of Seq for performance


polarker:

polarker:
The new ArraySeq might be even better I guess.

tdroxler:
[slick is using Vector](https://github.com/slick/slick/blob/56ea2b86064082b64684ea700871ec39a041a69e/slick/src/main/scala/slick/jdbc/StaticQuery.scala#L91-L104) for their api we use.

I'm not a collection expert, but should I give a try with `Vector` or try to override stuff to have `ArraySeq`?

polarker:
1. We'd better make it explicit, either `Vector` or `ArraySeq`
2. I prefer `ArraySeq`. This benchmark done years ago is still relevant: https://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#arrays-are-great

tdroxler:
:+1: I'll  give a try with `ArraySeq`

polarker:
If it's not possible, then we have to go with `Vector` I guess

tdroxler:
I did a quick try + benchmarks [here](https://github.com/alephium/explorer-backend/commit/af8bb35d) The difference is not huge, but overall it's a bit better that using `Seq`, this is only one query, if everything is `ArraySeq` it could indeed improve the overall thing.
Here are several runs:

```
DBBenchmark.getTransactionsByAddressSQL          thrpt       116.735          ops/s
DBBenchmark.getTransactionsByAddressSQLArraySeq  thrpt       167.559          ops/s
```
```
DBBenchmark.getTransactionsByAddressSQL          thrpt       169.956          ops/s
DBBenchmark.getTransactionsByAddressSQLArraySeq  thrpt       176.694          ops/s
```
```
DBBenchmark.getTransactionsByAddressSQL          thrpt       177.880          ops/s
DBBenchmark.getTransactionsByAddressSQLArraySeq  thrpt       188.152          ops/s
```
```
DBBenchmark.getTransactionsByAddressSQL          thrpt       125.048          ops/s
DBBenchmark.getTransactionsByAddressSQLArraySeq  thrpt       127.621          ops/s
```
```
DBBenchmark.getTransactionsByAddressSQL          thrpt       171.913          ops/s
DBBenchmark.getTransactionsByAddressSQLArraySeq  thrpt       174.093          ops/s
```
```
DBBenchmark.getTransactionsByAddressSQL          thrpt       170.659          ops/s
[...]

PR #327, opened 2022-08-18
Unconfirmed txs per address endpoint

### LINKED ISSUE #326: Unconfirmed txs per address endpoint
We want to be able to retrieve the unconfirmed txs for a given address. The purpose is with our different wallets, when we send a tx we gray it out until it's confirmed, but if you close the wallet during that period, the tx won't show when re-opening if it's still unconfirmed.

As discussed in [this PR](https://github.com/alephium/explorer-backend/pull/320) We miss a proper way to search the unconfirmed inputs for a user.

in the branch [feature/uinput-address](https://github.com/alephium/explorer-backend/tree/feature/uinput-address) I'm experimenting with a readable `UnlockScript`, which is part of the input.

`UnlockScript` can have 3 types:

- P2PKH: pay to public key hash, the classic one address makes a transfer
- P2MPKH: pay to multiple key hash: ALPH are sent from a multisig address.
- P2SH: pay to script hash: when a script is transfering hash.

In the branch I now have those 3 models implemented and the `Input` and `UInput` now have a field `unlockScript`
```
 "unlockScript": {
        "type": "p2pkh",
        "address": "14PqtYSSbwpUi2RJKUvv9yUwGafd6yHbEcke7ionuiE7w"
      }
```
```
 "unlockScript": {
        "type": "p2mpkh",
        "indexedAddresses": [{
           "address": "14PqtYSSbwpUi2RJKUvv9yUwGafd6yHbEcke7ionuiE7w",
           "index": 0
            },{
           "address": "152qtYSSbwpUi2RJKUvv9yUwGafd6yHbEcke7ion34TT",
           "index": 1
        }]
      }
```
```
 "unlockScript": {
        "type": "p2sh",
        "script": "<some_hex_string",
        "params": "<some_hex_string"
      }
```

Our wallets case is only interested in `p2pkh`, we could maybe add an endpoint:

`/addresses/<address>/p2pkh-unconfirmed-transactions`

WDYT @nop33

tdroxler:
I will still give a round of review of this PR, but I'm already opening it as I'm off the fridays.

resolves #326


app/src/main/scala/org/alephium/explorer/persistence/dao/UnconfirmedTxDao.scala
```diff
@@ -99,7 +103,35 @@ object UnconfirmedTxDao extends UnconfirmedTxDao {
...
+    run(for {
+      txHashes <- UInputSchema.table
+        .filter(_.p2pkhAddress === address)
+        .map(_.txHash)
+        .distinct
+        .result
+      txs     <- UnconfirmedTxSchema.table.filter(_.hash inSet txHashes).sortBy(_.lastSeen.desc).result
+      inputs  <- UInputSchema.table.filter(_.txHash inSet txHashes).result
+      outputs <- UOutputSchema.table.filter(_.txHash inSet txHashes).result
+    } yield {
```
simerplaha, line 128: Hey Thomas, just checking if we are working on reducing the number of concurrent queries dispatched? To avoid the incident from last week of high network IOps. 

From the looks of it, these queries could be a single query using `OUTER JOIN` etc 

Similarly in your last PR [here](https://github.com/alephium/explorer-backend/pull/320/files#diff-25723e7a7dcf656a9cecf4b53ce0e8ae1468880aaa4772ecfe503bd7b771886cR90-R95), these could also be written in a single query. Bringing this up now, in case these optimisations are being deferred to another issue.

simerplaha (commented):
Heya! Looks good. Big PR. First round done :-)

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomGetResult.scala
```diff
@@ -82,6 +82,18 @@ object CustomGetResult {
         .map(bytes =>
           deserialize[AVector[Token]](ByteString.fromArrayUnsafe(bytes)).toOption.get.toSeq)
 
+  @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
+  implicit val unlockScriptGetResult: GetResult[UnlockScript] =
+    (result: PositionedResult) =>
+      deserialize[UnlockScript](ByteString.fromArrayUnsafe(result.nextBytes())).toOption.get
```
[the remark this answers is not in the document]
tdroxler: ooh yes totally, when I wrote this code I told myself I need to go back on this before committing...seems I forgot :sweat_smile:

app/src/main/scala/org/alephium/explorer/persistence/dao/UnconfirmedTxDao.scala
tdroxler: I will need to benchmark all these, because one of our goal was to avoid heavy `JOIN` as they are slow. 
But I agree IOps definitively needs to be considered in the start now, thx for raising that :+1:

tdroxler: I'm thinking a bit more on the subject, for this case of unconfirmed transactions, where we don't do complicated queries and joining. We could maybe actually store the serialization of the tx as a byte array in the `UnconfirmedTxSchema`, this would look a bit like a key value store. 

Any opinion on this?

tdroxler: I gave a try [here](https://github.com/alephium/explorer-backend/commit/91974b22ef671120065a999224db4e39da8ad7c8)

simerplaha: > I gave a try [here](https://github.com/alephium/explorer-backend/commit/91974b22ef671120065a999224db4e39da8ad7c8)

Nice 👍🏼. Left a small comment. 

>  We could maybe actually store the serialization of the tx as a byte array in the `UnconfirmedTxSchema`, this would look a bit like a key value store.
>
> Any opinion on this?

Oh we do something similar with `AppStateSchema` right. 

My *opinion*, I'm not sure. But it be nice if the data we create is laid out the way the database we use likes it laid out as it's optimised for it. And incase we do wanna query the table in the future for some admin UI and display analytics. But thats just *opinion* :) I'm sure there are good reasons for it to be a key-value table.

tdroxler: yes I agree we still need to keep the data "queryable", the idea is not to have dynamic values, as for a tx_id the data is immutable. In this  `UnconfirmedTx` case, we still store all data the sql way, so we can make more complex queries in the future, but storing the full transaction at the same time allows to avoid JOINS when not necessary....But I'm still not sure too  ....maybe it's not worth it for the `UnconfirmedTx` as the table stays small.

tdroxler: In the end I went first with a solution following what we have in our `TransactionQueries`, using only raw sql to improve perf.
I'm not sure we can nicely joins the 3 tables, as inputs and outputs are really 2 different things and don't need to be joined.

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomGetResult.scala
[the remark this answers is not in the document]
tdroxler: done

app/src/test/scala/org/alephium/explorer/Generators.scala
```diff
@@ -132,10 +133,34 @@ object Generators {
     key  <- hashGen
   } yield protocolApi.OutputRef(hint, key)
 
+  lazy val keypairGen: Gen[(PrivateKey, PublicKey)] =
+    Gen.const(()).map(_ => SignatureSchema.secureGeneratePriPub())
```
simerplaha, line 137: Just checking the use-case for a `lazy val` here. A normal `val` should be good here right?

app/src/main/scala/org/alephium/explorer/persistence/dao/UnconfirmedTxDao.scala
simerplaha: Oh it looked like three queries (3 DB trips) running on the same input so a join should be possible. But maybe I'm overlooking something. Please mark this as resolved if you are ok with it.

tdroxler: the problem I see:
first we join the tx table with the input on the `tx_hash`, let say we have 3 inputs for one tx: the output rows will be:
```
tx, in1
tx, in2,
tx, in3
```
now we need to join to get the outputs also on the `tx_hash`, let say 2 outputs:
```
tx, in1, out1
tx, in1, out2
tx, in2, out1
tx, in2, out2
tx, in3, out1
tx, in3, out3
```
Am I right?

We unnecessary return multiple time the inputs and outputs, while we only want 1 tx, 3 inputs and 2 outputs. Imagine a tx with 100 inputs and 100 outputs.

But maybe I miss something? or maybe it's not an issue compared to querying 3 times the DB?

app/src/test/scala/org/alephium/explorer/Generators.scala
tdroxler: totally.....I copy pasted this from `alephium` and didn't check further :see_no_evil:

app/src/main/scala/org/alephium/explorer/persistence/dao/UnconfirmedTxDao.scala
simerplaha: You are right. It would return duplicate rows. Which is ok. But why would there be a need to read 100 inputs and 100 outputs in a single query anyway? Following your example, that would be 10,000 rows? We shouldn't be reading that much data in a single query anyway unless it's for aggregation/analytics which can be done within the query and deferred to the DB. Do we have a UI that displays that much data? I'm trying to look for where `UnconfirmedTxDao.list` and `listByP2PKHAddress` are being called, I couldn't find anything.

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -323,9 +323,14 @@ object BlockFlowClient {
     )
 
   private def inputToUInput(input: api.model.AssetInput): UInput = {
+    val unlockScript = input.toProtocol() match {
+      case Left(error)  => throw new ExplorerError.InvalidProtocolInput(error)
+      case Right(value) => value.unlockScript
+    }
```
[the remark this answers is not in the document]
tdroxler: agree on the function to remove the duplicate.

For the `Try`, we would need to propagate the `Try` up to the top and we endup with some `Seq[Try[A]]`, it seems to bring a lot of complexity while throwing an error is also failing the Future right?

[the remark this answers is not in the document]
simerplaha: Yep you are right. `throw` will also fail the `Future`. Just a syntax thing because if I have another function that is not within the `Future`, I will also need to learn what's in this function so that new function can handle the `throw`. If it's returning `Try`, I wouldn't need to learn about this function.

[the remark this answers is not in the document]
tdroxler: oh good point, I could indeed return a `Try`, but throw from where I use that function

app/src/main/scala/org/alephium/explorer/persistence/dao/UnconfirmedTxDao.scala
simerplaha: Thomas I will leave this to your judgement. I'm sure this is well thought out. Don't think I'm being helpful here, I'm just raising more questions. I can always give it a performance check later, if possible.

tdroxler: if someone is requesting the details of such a tx, we have to return every inputs and outputs for that given tx, we can't afford to skip some of them otherwise if you sum up everything you will think that you miss some ALPH. Knowing that all those transactions endpoints are also called from the wallet, it would be a problem.
And yes in the explorer we will display all the inputs and outputs in the transaction page

tdroxler: your questions are always good and helps me a lot improving the perfs, in this particular case it gave me the opportunity to really think of a "one query thing". I just couldn't find a good one fitting our requirement, maybe there is....

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomJdbcTypes.scala
```diff
@@ -97,18 +97,34 @@ object CustomJdbcTypes {
...
-  @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
   implicit val seqByteStringType: JdbcType[Seq[ByteString]] =
     MappedJdbcType.base[Seq[ByteString], Array[Byte]](
       byteStrings => serialize(AVector.unsafe(byteStrings.toArray)).toArray,
       bytes =>
-        deserialize[AVector[ByteString]](ByteString.fromArrayUnsafe(bytes)).toOption.get.toSeq
+        deserialize[AVector[ByteString]](ByteString.fromArrayUnsafe(bytes)) match {
+          case Left(error)  => throw error
+          case Right(value) => value.toSeq
+      }
```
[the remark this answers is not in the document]
tdroxler: yes you are totally right, it's not the first time I shamefully wrote `toSeq`. 

In the explorer backend we deal with `Seq` (or [maybe IndexedSeq](https://github.com/alephium/explorer-backend/issues/148)) as slick is returning such collection, but when we are dealing with `alephium` model, we get the `AVector`. 
In the particular case here, I will have a look if we can deserialize a serialized `AVector` directly into a `Seq`, but it won't solve the `toSeq` inside the `BlockFlowClient` as we are receiving the models directly from `alephium`.
One solution to dig would be to use `AVector` everywhere, but I don't know if we can override slick to return an `AVector` instead of `Seq`, knowing slick my gut feeling is no, but I'll have a look, the goal is not to to do an `AVector.from(seq)` everywhere either....

tdroxler:
I added few commits base on internal discussion with @polarker , I think we are good for another review.

polarker (approved):
LGTM, nothing more to complain 👍

simerplaha (approved):
🚀

Issue #324, opened 2022-08-16
Disable `inline` logs from build log output 


simerplaha:
Inline logs fill up the CI build log output.

We should probably disable logging for all builds and enable manually when debugging?

tdroxler:
mmmm can't we just disable inline logs? 

When something is flaky it's convenient to have the logs right away, as it doesn't always fail.

simerplaha:
> mmmm can't we just disable inline logs?

Yep. That's exactly what the issue is suggesting. Titled: Disable `inline` logs.

tdroxler:
oh I thought you were proposing to remove ALL logs :sweat_smile:

PR #329, opened 2022-08-25
Disable inline logs

tdroxler:
resolves https://github.com/alephium/explorer-backend/issues/324

is it enough for you @simerplaha ?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,8 +91,11 @@ val inliningOptions =
     "-opt-warnings",
     "-opt:l:inline",
     "-opt-inline-from:org.alephium.explorer.**",
+    //Uncomment to debug inlining
+    /*
     "-Yopt-log-inline",
     "_"
+   */
   )
 
 val commonSettings = Seq(


tdroxler:
@simerplaha `unidoc` is failing in [this line](https://github.com/alephium/explorer-backend/blob/cbe07e5f533ea805fa22715dce64fcbbe713b11b/build.sbt#L95)  what's its purpose?

simerplaha:
The `"_"` means to print all. Documentation [here](https://docs.scala-lang.org/overviews/compiler-options/index.html). 

> -Yopt-log-inline PACKAGE/CLASS.METHOD
Print a summary of inliner activity; _ to print all, prefix match to select.

tdroxler:
> The `"_"` means to print all. Documentation [here](https://docs.scala-lang.org/overviews/compiler-options/index.html).
> 
> > -Yopt-log-inline PACKAGE/CLASS.METHOD
> > Print a summary of inliner activity; _ to print all, prefix match to select.

ok so we can comment that line too, thxxx

simerplaha (approved):
🚀

PR #332, opened 2022-08-31
#328: Query for `getLocalMaxTimestamp`

simerplaha:
- Resolves #328. 
- Ran the test-case in this PR over 100s of iterations. The test also checks for same output as existing code. So it should cover all existing cases. 

Started off with typed query [here](https://github.com/alephium/explorer-backend/blob/328_getLocalMaxTimestamp_2/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala#L342) hoping to reuse existing queries to speed things up. But the generated `UNION` queries looked like following so switched to SQL. 

## TODO
- Remove deprecated code. See TODO in code. 
- Benchmarks. Please let me if you'd like this done.

## Generated query
<details>
  <summary>Generated query</summary>

  ```sql
(select x2.x3 as x4, x2.x5 as x6
 from ((select x7.x8 as x3, x7.x9 as x5
        from ((select x10.x11 as x8, x10.x12 as x9
               from ((select x13.x14 as x11, x13.x15 as x12
                      from ((select x16.x17 as x14, x16.x18 as x15
                             from ((select x19.x20 as x17, x19.x21 as x18
                                    from ((select x22.x23 as x20, x22.x24 as x21
                                           from ((select "height" as x23, max("block_timestamp") as x24
                                                  from "block_headers"
                                                  where (("height" = (select max("height")
                                                                      from "block_headers"
                                                                      where ("chain_from" = 0)
                                                                        and ("chain_to" = 0))) and ("chain_from" = 0))
                                                    and ("chain_to" = 0)
                                                  group by "height")
                                                 union all
                                                 (select "height" as x23, max("block_timestamp") as x24
                                                  from "block_headers"
                                                  where (("height" = (select max("height")
                                                                      from "block_headers"
                                                                      where ("chain_from" = 0)
                                                                        and ("chain_to" = 1))) and ("chain_from" = 0))
                                                    and ("chain_to" = 1)
                                                  group by "height")) x22)
                                          union all
                                          (select "height" as x20, max("block_timestamp") as x21
                                           from "block_headers"
                                           where (("height" = (select max("height")
                                                               from "block_headers"
                                                               where ("chain_from" = 0) and ("chain_to" = 2))) and
                                                  ("chain_from" = 0))
                                             and ("chain_to" = 2)
                                           group by "height")) x19)
                                   union all
                                   (select "height" as x17, max("block_timestamp") as x18
                                    from "block_headers"
                                    where (("height" = (select max("height")
[...]

PR #334, opened 2022-09-01
#333: Fetch data required by `BlockFlowSyncService.init()` in single query

### LINKED ISSUE #333: Reduce queries dispatched by `BlockFlowSyncService.init()`
[BlockFlowSyncService.init()](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L151) dispatches 16 queries concurrently to fetch `maxHeight` when `group-num = 4`. 

This data can be fetched in a single query reducing DB trip to 1.

The data can then be distributed to multiple threads locally for concurrent processing.

simerplaha:
- Resolves #333
- This is PR is stacked on #332 (needed the functions implemented there).

## TODO

[`BlockFlowSyncService.init()`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L151) has no existing unit tests. So can't really compare the function's output against this change.  

Please let me know if you'd like test-cases for `BlockFlowSyncService.init()` implemented.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
@@ -372,4 +372,57 @@ object BlockQueries extends StrictLogging {
       unitPConv  = parameters
     ).as[Option[(TimeStamp, Height)]].oneOrNone
   }
+
+  /**
+    * Fetches maximum `Height` for each `GroupIndex` pair.
+    *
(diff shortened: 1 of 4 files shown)


tdroxler (approved):
LGTM, same remark as https://github.com/alephium/explorer-backend/pull/332#pullrequestreview-1093587163

PR #341, opened 2022-09-05
#335: Fix daylight savings test issue

### LINKED ISSUE #335: `SchedulerSpec`: Test failed `should return today's duration`
[Build failed](https://github.com/alephium/explorer-backend/runs/8149322870?check_suite_focus=true#step:5:1622). Looks like a test only issue.

``` 
[info] SchedulerSpec:
[info] scheduleTime
[info]   when time is now
[info]   - should return now/current/immediate duration
[info]   when time is in the future
[info]   - should return today's duration *** FAILED ***
[info]     TestFailedException was thrown during property evaluation.
[info]       Message: 9 did not equal 10
[info]       Location: (SchedulerSpec.scala:76)
[info]       Occurred when passed generated values (
[info]         arg0 = America/Santiago
[info]       )
[info]     Init Seed: 8851152591385333189
```

simerplaha:
- Resolves #335
- The test was failing for timezone `America/Santiago`, it went in daylight saving recently ( -3 hours) making the test-case invalid. I've restricted that test to run for only known timezones (Sydney, Switzerland) so its more predictable.
- Also added more seconds to other test-cases, just incase the build is slow.

### DIFF
--- a/app/src/test/scala/org/alephium/explorer/util/SchedulerSpec.scala
+++ b/app/src/test/scala/org/alephium/explorer/util/SchedulerSpec.scala
@@ -50,7 +50,7 @@ class SchedulerSpec
         forAll(Gen.oneOf(zoneIds)) { zoneId =>
           //Time is now! Expect hours and minutes to be zero. Add 2 seconds to account for execution time.
           val timeLeft =
-            Scheduler.scheduleTime(ZonedDateTime.now(zoneId).plusSeconds(2), "test-scheduler")
+            Scheduler.scheduleTime(ZonedDateTime.now(zoneId).plusSeconds(20), "test-scheduler")
 
           timeLeft.toHours is 0
           timeLeft.toMinutes is 0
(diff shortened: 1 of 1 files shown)


app/src/test/scala/org/alephium/explorer/util/SchedulerSpec.scala
```diff
@@ -83,16 +83,25 @@ class SchedulerSpec
...
             .toHours is 23
+        }
+
+        /**
+          * <a href="https://github.com/alephium/explorer-backend/issues/335">#335</a>:
+          * The following test is invalid for timezones in daylight savings.
+          *
+          * Restricting this test to run for known timezones only (Sydney & Switzerland).
+          * */
+        Seq(ZoneId.of("Australia/Sydney"), ZoneId.of("CET")) foreach { zoneId =>
```
tdroxler, line 104: what will happen when Switzerland goes in light saving mode too? (+/- 1 hour)?

simerplaha: Same goes with Sydney. We go in daylight savings as well (+/- 1 hour). In the comment I mentioned at least its predictable for countries we care about. This is a test-case only issue, the scheduling itself will work as expected as proven by other test-cases. The whole daylight savings and how it's computed is something I don't understand. 

If you'd like the test-case work for all daylight savings cases, I can read up on it?

tdroxler: nope all good for me, thx for the explanation

tdroxler (approved):

PR #342, opened 2022-09-05
Fix/input update

tdroxler:
In a recent commit, we added directly the `output_ref_address`
to the `Input` when available through the `UnlockScript`.
The `InputUpdateQuery` was then failing as we were searching for
the empty addresses. We are now looking for empty amount, an index is
added and we can see that the query planner is using our new
`inputs_output_ref_amount_null_idx` index

```
EXPLAIN UPDATE inputs
SET
  output_ref_address = outputs.address,
  output_ref_amount = outputs.amount,
  output_ref_tokens = outputs.tokens
FROM outputs
WHERE inputs.output_ref_key = outputs.key
AND inputs.output_ref_amount IS NULL
RETURNING outputs.address, outputs.tokens, inputs.tx_hash, inputs.block_hash, inputs.block_timestamp, inputs.tx_order, inputs.main_chain
                                                 QUERY PLAN
-------------------------------------------------------------------------------------------------------------
 Update on inputs  (cost=0.69..16.73 rows=1 width=100)
   ->  Nested Loop  (cost=0.69..16.73 rows=1 width=100)
         ->  Index Scan using inputs_output_ref_amount_null_idx on inputs  (cost=0.12..8.14 rows=1 width=39)
               Index Cond: (output_ref_amount IS NULL)
         ->  Index Scan using outputs_pk on outputs  (cost=0.56..8.58 rows=1 width=127)
               Index Cond: (key = inputs.output_ref_key)
```

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala
@@ -96,6 +96,7 @@ object DBInitializer extends StrictLogging {
       _ <- OutputSchema.createMainChainIndex()
       _ <- TransactionPerAddressSchema.createMainChainIndex()
       _ <- OutputSchema.createNonSpentIndex()
+      _ <- InputSchema.createOutputRefAddressNullIndex()
     } yield ())
   }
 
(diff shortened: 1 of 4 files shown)


simerplaha (approved):
Awesome 🚀 🚀

PR #345, opened 2022-09-10
Create nobody home folder

sven-hash:
This error `at java.io.FileNotFoundException: /nonexistent/.explorer/logs/explorer-info.log (No such file or directory)` appears because no homefolder exists for user nobody


tdroxler (approved):
thx :rocket:

PR #346, opened 2022-09-12
Fix token transactions queries

tdroxler:
raw queries was using a wrongly named column....and there was no test :sweat_smile:


simerplaha (approved):
🚀

PR #349, opened 2022-09-14
Move all the collections to use only `AVector`

tdroxler:
Resolves https://github.com/alephium/explorer-backend/issues/148 
I kept the first commit with the `ArraySeq`conversion just in case, but better to read the full diff rather than commit per commit.
Having `AVector` as an `IterableOnce` helped a looooot, thx again @polarker . Too bad we can't have `Iterable` for few little things.

I had to add a [BuildFrom[AVector] ](https://github.com/alephium/explorer-backend/blob/d073c05a33ec52f93a0209ea683d4e8a963f1c5b/app/src/main/scala/org/alephium/explorer/RichAVector.scala#L25-L34)in order tu be able to use [Future.sequence](https://www.scala-lang.org/api/current/scala/concurrent/Future$.html#sequence[A,CC[X]%3C:IterableOnce[X],To](in:CC[scala.concurrent.Future[A]])(implicitbf:scala.collection.BuildFrom[CC[scala.concurrent.Future[A]],A,To],implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[To]), maybe we could add this to `AVector`

Few other things we could add, but not that important as it's only used 1-2 times in the code:
* [asJava](https://github.com/alephium/explorer-backend/blob/d073c05a33ec52f93a0209ea683d4e8a963f1c5b/app/src/main/scala/org/alephium/explorer/cache/CaffeineAsyncCache.scala#L83-L85)
* [zip](https://github.com/alephium/explorer-backend/blob/d073c05a33ec52f93a0209ea683d4e8a963f1c5b/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala#L422-L423) (we only have `zipWithIndex`)
* [splitAtUpTo](https://github.com/alephium/explorer-backend/blob/d073c05a33ec52f93a0209ea683d4e8a963f1c5b/app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala#L134-L136) the equivalent of [this](https://www.scala-lang.org/api/current/scala/collection/Seq.html#splitAt(n:Int):(C,C)) which is equivalent to (but possibly more efficient than) (c take n, c drop n) according to the scala doc
* a [groupByToAvector](https://github.com/alephium/explorer-backend/blob/d073c05a33ec52f93a0209ea683d4e8a963f1c5b/app/src/main/scala/org/alephium/explorer/persistence/queries/OutputQueries.scala#L264-L272)  That something I always wanted with any collection, using `groupBy` is super useful, but most of the time we later want an `AVector[(K,V)]` rather than a `Map[K,V]`


polarker:
Just to note: I am totally fine with ArraySeq, or even Vector, as long as they are used explicitly in the codebase.

It's up to you guys to decide if it's worth to use `AVector` in the codebase.

simerplaha (approved):
I don't know the motivation behind this change - “Let’s use `AVector` for everything”.

`AVector` is also a partially defined collection, its needs a few [implementations](https://docs.scala-lang.org/overviews/core/custom-collections.html) to be complete. 

I’m sure this is something you guys want for a reason.

Don’t see any bugs. Approved.

polarker:
> I don't know the motivation behind this change - “Let’s use `AVector` for everything”.
> 
> `AVector` is also a partially defined collection, its needs a few [implementations](https://docs.scala-lang.org/overviews/core/custom-collections.html) to be complete.
> 
> I’m sure this is something you guys want for a reason.
> 
> Don’t see any bugs. Approved.

There are some challenges to make it fully compatible with Scala's collection while keeping the same performance&simplicity as I wished for full node. In this PR, Thomas was trying to see how far we could go with `AVector`.

Thank @tdroxler for the great work, especially the improvement points 👍  Given what @simerplaha said, my personal feeling is that we should go with explicit `ArraySeq`. @tdroxler WDYT?

tdroxler:
>Given what @simerplaha said, my personal feeling is that we should go with explicit ArraySeq. @tdroxler WDYT?

Yes I feel the same, using `AVector` was useful too keep the same type as received from the full node API, but it lacks just few little things to make it convenient to use in the explorer-backend, especially the communication with others libraries.

@simerplaha the main idea was to have a more efficient collection than `Seq` and indeed using `AVector` was first a test to see how far we could go. I could go up to the end, but as we see it's probably not suitable for now. 

Hopefully I kept the commit were I migrated to `ArraySeq`, I'll go back to it and clean it a bit.

simerplaha:
Wow! A lot of work done here. You are a machine 💪🏼 

I will give it another go tomorrow. It looks neat 🚀 🚀

app/src/main/scala/org/alephium/explorer/RichAVector.scala
```diff
@@ -0,0 +1,30 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+package org.alephium.explorer
+
+import scala.collection.immutable.ArraySeq
+
+import org.alephium.util.AVector
+
+//TODO Add this to `org.alephium.util.AVector`
```
polarker, line 23: We don't add this on purpose because this implicit conversion needs to allocate a new array. It's expensive for large vectors.

tdroxler: I was not thinking of an implicit conversion, but just adding the `toArraySeq` function alongside [the existing toSeq](https://github.com/alephium/alephium/blob/d8b20ac2bb4578dc64aecb9e4d3a03063dd0c4b1/util/src/main/scala/org/alephium/util/AVector.scala#L650-L652)

polarker: I see

Issue #270, opened 2022-06-16
Migrate to vertx


tdroxler:
to get rid of akka-http and follow what we have in the full node.

simerplaha:
To save duplicating work, maybe a part of #256 can be implemented in this issue or we can implement this after #256. 

We kinda should decouple explorer APIs from the http-server we use. Vertx, Akka-http or whatever should only know about a single function it needs to call per route.  Currently all routes have to know about everything it needs to execute a request - `ExecutionContext`, `DatabaseConfig`, `*Cache`, `*Service` etc and are therefore stateful.

We need `ExplorerAPI` that exposes only the public APIs to be used by the http-server. Makes testing easier, test-cases in `ApplicationSpec/ExplorerSpec` will be typed.

tdroxler:
good idea to separate a bit. I need to check if we can test an [Endpoint](https://github.com/alephium/explorer-backend/blob/abe41362cf13d2f7bf2face5a251d0e060eaadb6/app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala#L35-L39) without it's [serverLogic](https://github.com/alephium/explorer-backend/blob/abe41362cf13d2f7bf2face5a251d0e060eaadb6/app/src/main/scala/org/alephium/explorer/web/AddressServer.scala#L42-L47)

PR #357, opened 2022-09-21
Update common to `1.8.5-rc8`

tdroxler:
Creating this as a draft as there is still one thing I need to investigate. For some reason if[ I don't manually add manually "type" field in some case class ](https://github.com/alephium/explorer-backend/blob/738ef7b9aebda4ac44dce357c02e35fed50a8731/app/src/main/scala/org/alephium/explorer/api/model/Output.scala#L40-L51) the `openapi.json` then create two types: `AssetOutput` and `AssetOutput1`. This is because some endpoints use directly `AssetOutput` and some other through the `sealed trait Output`. I need to check tapir and upickle to see what changed.

This all PR is the first step to integrate the smart contract events, I thought I could directly start with them, but the changes are quite involving.
BTW I love all the new types, instead of just using `Hash`.

### DIFF
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -464,7 +464,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/ConfirmedTransaction"
+                  "$ref": "#/components/schemas/Transaction"
                 }
               }
             }
@@ -883,7 +883,7 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/UnconfirmedTransaction"
+                    "$ref": "#/components/schemas/TransactionLike"
                   }
                 }
               }
@@ -1717,7 +1717,7 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/UnconfirmedTransaction"
+                    "$ref": "#/components/schemas/TransactionLike"
                   }
                 }
               }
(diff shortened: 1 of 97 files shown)


polarker (approved):
I reviewed it quickly. LGTM. We want to make a RC for test.

simerplaha (approved):
🚀 🚀 🚀

polarker:
As @Lbqds mentioned, we are using different versions of upickle.

I did not upgrade upickle to `2.0.0` due to some bugs. It might be relevant to your issue too. @tdroxler

tdroxler:
@polarker I saw you released `v1.9.0-rc0` based on your branch. Should I close this PR?

tdroxler:
I pushed 2 new commits, super headache stuff here.
as we discussed internally, [tapir 1.0.6 broke something](https://github.com/softwaremill/tapir/issues/2461), if I'm using `1.0.5` on master it's all good, but if I do it on this PR, I still have the weird issue, I tried to exclude `tapir` from our `alephium` lib, but it was still failing. So in the first commit I went another way to avoid having the bug described above.
For the second commit, I downgraded `upickle` to `1.6.0` and remove `akka-http-upickle

`ExplorerSpec` was failing, it appears that there is an issue between akka-http
marshalling/unmarshalling and upickle.

I first thought it was due to `akka-http-upickle` as it was not using
`upickle-2.0.0`, so I wrote our own version of that lib (which is super
small), but the problem still occurs, don't know why :/

That could be a good reason to tackle https://github.com/alephium/explorer-backend/issues/270

Issue #356, opened 2022-09-19
Services started before `startSelfOnce` is invoked & `Scheduler` is not closed


simerplaha:
In [ExplorerState.ReadWrite](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/ExplorerState.scala#L117) the `Scheduler` is not closed on termination and is always started regardless `startSelfOnce` is called or not. 

[DatabaseConfig[PostgresProfile]](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/Database.scala#L31) is provided pre-started to this  `Database` class, its not fully managed by this class. So it's also started regardless `startSelfOnce` is called or not.

Be nice if this stuff was strongly typed so this gets caught during compile time.

polarker:
Curious how type would solve this issue?

Imagine the case that a developer initialize a class without realizing that it should be closed. Will type be helpful in this situation?

simerplaha:
> Imagine the case that a developer initialize a class without realizing that it should be closed. Will type be helpful in this situation?

For any generic class I don't have idea on how to. That would be a Rust thing I guess or capture-checking in Scala3. But for something around the `Service` trait, I think I would've made up some data structure that allows me to expose those `start` and `stop/close` functions/functionality only in specific states. So only an `Initialised` service can be started and only a `Started` service can be stopped.

```scala
sealed trait ServiceStatus
case class Initialised[T <: Service](service: T) extends ServiceStatus
case class Started[T <: Service](service: T) extends ServiceStatus
case class Terminated[T <: Service](service: T) extends ServiceStatus
```

Personally I'd also use some version of resource management using functions instead of the `Service` trait so things are composable, immutable and we don't need to manually define parent child hierarchy.

polarker:
> But for something around the Service trait...

It's the same IMO. The issue is that Vertx is not wrapped as a service, I did not do it for Database too.

Once it's clear what should be treated as services, then there are different ways to manage the lifecycle of those services. No matter what approach you go, you will need to manage the dependencies in some ways.

The tricky part in those services that are initialized inside another service. For such cases, the compiler cannot know if you really want just a service or you want to use a service with a specific status, so type system cannot help either.

tdroxler:
We didn't have any issue with scheduler since 2022, closing

PR #361, opened 2022-09-29
Migrate to `vertx`

tdroxler:
I tried to have a readable diff, I redo what I did in the node, keeping same wording as akka-http-testkit.

Unfortunatelly with `vertx` we have to start the server: `server.listen` in order to test the routes, we can't do as in akka.

So now for all the `web.*ServerSpec` I start once the server on top of the file and all test are using the same instance.

Resolves:  #270


tdroxler:
we are still using `akka` [ to stop the all system ](https://github.com/alephium/explorer-backend/blob/6189c596cbe991d30de5e4c4b6adf0d8d3d1a4b2/app/src/main/scala/org/alephium/explorer/SyncServices.scala#L101) for sure we can do differently.

We also use `ByteString` but this could be replace by a simple `Array[Byte]`

app/src/main/scala/org/alephium/explorer/ExploreHttpServer.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  /** Stop AkkaHttp server and terminates ActorSystem */
+  def stopSelfOnce(): Future[Unit] = {
+    for {
+      binding <- httpBindingPromise.future
+      _       <- binding.close().asScala
+    } yield {
+      logger.info(s"http unbound")
+    }
+  }
```
simerplaha, line 85: Think there is a need close the Vertx instance here with `vertx.close()`? Without it the Thread is held

> [WARN] [10/03/2022 15:35:46.749] [default-akka.actor.internal-dispatcher-2] [CoordinatedShutdown(akka://default)] Coordinated shutdown phase [before-actor-system-terminate] timed out after 5000 milliseconds
2022-10-03 15:35:50,078 [-dispatcher-4] INFO  c.z.h.HikariDataSource                 - db.db - Shutdown completed.
2022-10-03 15:35:50,084 [-dispatcher-5] INFO  o.a.e.ExplorerHttpServer               - http unbound

With `vertx.close()` it properly exists

> 2022-10-03 15:37:41,909 [-dispatcher-6] INFO  c.z.h.HikariDataSource                 - db.db - Shutdown completed.
2022-10-03 15:37:41,936 [dispatcher-11] INFO  o.a.e.ExplorerHttpServer               - http unbound
Process finished with exit code 0

app/src/main/scala/org/alephium/explorer/ExploreHttpServer.scala
```diff
@@ -0,0 +1,88 @@
...
+
+/** Stores AkkaHttp related instances created on boot-up */
+@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+class ExplorerHttpServer(host: String, port: Int, val routes: ArraySeq[Router => Route])(
+    implicit val executionContext: ExecutionContext
+) extends Service
+    with StrictLogging {
+
+  private val vertx  = Vertx.vertx()
+  private val router = Router.router(vertx)
```
simerplaha, line 39: The issue #356 might be valid here? `vertx` is being created without calling `startSelfOnce`.

tdroxler: nice catch, I'll have a look

app/src/main/scala/org/alephium/explorer/ExploreHttpServer.scala
tdroxler: oh nice catch, I didn't thought vertx  needed to be close!

app/src/main/scala/org/alephium/explorer/ExploreHttpServer.scala
tdroxler: done

app/src/main/scala/org/alephium/explorer/ExploreHttpServer.scala
tdroxler: done

simerplaha (approved):
🚀 🚀 🚀

PR #362, opened 2022-10-05
Sync performance

simerplaha:
@tdroxler, this draft is for early feedback to see if you are happy with the direction this is taking.

Not at all ready for merge :)

Code comments should explain things.


app/src/main/scala/org/alephium/explorer/service/sync/SimersSyncPlayground.scala
```diff
@@ -0,0 +1,312 @@
...
+                                            groupSetting: GroupSetting,
+                                            cache: BlockCache): Future[Unit] =
+    if (blocks.isEmpty)
+      Future.unit
+    else
+      for {
+        outOfSyncBlocks <- findOutOfSyncBlocks(blocks) //find out-of-sync blocks
+        missingBlocks <- executeActions(outOfSyncBlocks) //persist out of sync blocks
+        fetchedMissingBlocks <- client.fetchMissingBlocks(missingBlocks) //updateMainChain returns missing blocks
+        result <- insert(fetchedMissingBlocks) //persist those missing blocks
```
tdroxler, line 84: I can't find the place where you actually insert the `blocks`

tdroxler: This is actually an issue for the `findOutOfSyncBlocks` functions.
Let say we have 
```
B0 -> B1 
```
already in DB and we need to insert 
```
B2->B3->B4
```
when calling `findOutOfSyncBlocks` on the 3 new blocks, for `B2` it's fine as it's parent `B1` is alreay in db, even if it was not, it's correct to fetch it. But as we don't (at least I can't see it) insert `B2`, when calling `findOutOfSyncBlock` on `B3`, it won't find it's parent `B2` as it wasn't inserted, we will thus refetch it for nothing.
Correct me if I miss something.

tdroxler (commented):
I see some nice idea, but I'm afraid to make too many changes at once, as this is the most critical part of the explorer-backend.

I know this is a playground here, let's discuss on Slack on how to proceed. I probably need to come with a clear plan on how to improve this part.

app/src/main/scala/org/alephium/explorer/service/sync/SimersSyncPlayground.scala
simerplaha: > I can't find the place where you actually insert the blocks
The insert happens in one place [here](https://github.com/alephium/explorer-backend/pull/362/files#diff-6e6921810a0484ff6cfb7cb484079a5f39b570b61ac07573f4ad79549134dcedR167).

> B2->B3->B4

You are totally right. This is not something I've gotten around to yet, to avoid double fetching, but it has a simple solution. 

This PR was like my last attempt at understanding the sync process. Was having trouble understanding whats going on under the hood. Still unsure if I fully understand it. 

Totally agree with you. Chat on Slack.

PR #363, opened 2022-10-05
Replace delayed `Future` execution

simerplaha:
`fetchLocalTs` and `fetchRemoteTs` are not conditional and are always executed. Updated `fetchAndBuildTimeStampRange` to be a pure function. 

Replaced the following 

```scala
def fetchAndBuildTimeStampRange(
      step: Duration,
      backStep: Duration,
      fetchLocalTs:  => Future[Option[(TimeStamp, Int)]],
      fetchRemoteTs: => Future[Option[(TimeStamp, Int)]]
  )
```

with 

```scala
def fetchAndBuildTimeStampRange(
      step: Duration,
      backStep: Duration,
      localTs: Option[(TimeStamp, Int)],
      remoteTs: Option[(TimeStamp, Int)]
)
```

## TODO

[This test case](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/service/BlockFlowSyncServiceSpec.scala#L50) does not perform any IO operation but is creating a database connection and dropping/creating tables because the class extends `DatabaseFixtureForAll`. Not sure how you'd like this handled.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
@@ -178,9 +178,11 @@ case object BlockFlowSyncService extends StrictLogging {
       implicit ec: ExecutionContext,
       dc: DatabaseConfig[PostgresProfile],
       blockFlowClient: BlockFlowClient,
-      groupSetting: GroupSetting): Future[(ArraySeq[(TimeStamp, TimeStamp)], Int)] = {
-    fetchAndBuildTimeStampRange(step, backStep, getLocalMaxTimestamp(), getRemoteMaxTimestamp())
-  }
+      groupSetting: GroupSetting): Future[(ArraySeq[(TimeStamp, TimeStamp)], Int)] =
+    for {
+      localTs  <- getLocalMaxTimestamp()
+      remoteTs <- getRemoteMaxTimestamp()
+    } yield fetchAndBuildTimeStampRange(step, backStep, localTs, remoteTs)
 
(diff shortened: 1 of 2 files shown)


simerplaha:
@tdroxler, are these failing test-cases relevant to this PR? Doesn't seem like it 🤔

tdroxler:
>[This test case](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/service/BlockFlowSyncServiceSpec.scala#L50) does not perform any IO operation but is creating a database connection and dropping/creating tables because the class extends DatabaseFixtureForAll. Not sure how you'd like this handled.

We could definitively change that, we could maybe extract the function you changed in this PR into `TimeUtil`? or create a new `*Util` file?

tdroxler (approved):

tdroxler:
> @tdroxler, are these failing test-cases relevant to this PR? Doesn't seem like it thinking

yeah, I downloaded the logs and it's not related, I re-launched the test and it's good now.

simerplaha:
> > [This test case](https://github.com/alephium/explorer-backend/blob/master/app/src/test/scala/org/alephium/explorer/service/BlockFlowSyncServiceSpec.scala#L50) does not perform any IO operation but is creating a database connection and dropping/creating tables because the class extends DatabaseFixtureForAll. Not sure how you'd like this handled.
> 
> We could definitively change that, we could maybe extract the function you changed in this PR into `TimeUtil`? or create a new `*Util` file?

Sounds good. Will do this change in another PR because that `sys.exit(0)` might need to handled differently if it's moved to `TimeUtil`.

Issue #358, opened 2022-09-22
Remove double reading a `block` in `BlockFlowSyncService.insert`


simerplaha:
Function call to [`updateMainMain`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L279) should process the already read `parentBlock` instead  of fetching the same block again from the database in [`BlockDao.updateMainChain`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala#L113).

The asserts [`assert(block.chainFrom == chainFrom && block.chainTo == chainTo)`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala#L116) and the check for [`!block.mainChain`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala#L115) are also not required (by the current `sync` thread) as it's already confirmed by [`BlockFlowSyncService.insert`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L275).

Parent issue: #350

@tdroxler please confirm, is this a valid issue?

tdroxler:
definitively, I think the idea was to make `BlockDao.updateMainChainAction` safe and usable by everyone, but it seems it's used just by the `BlockFlowSyncService` so we could definitively improve few things.

simerplaha:
Temporarily closing this. PR #369 has changed the story here. 

Will reopen if needed as sync optimisations (#343) progress.

Issue #328, opened 2022-08-24
Reduce number of queries dispatched by `getLocalMaxTimestamp` 


simerplaha:
One of the first queries executed by sync is [`getLocalMaxTimestamp`](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L181) to calculate maximum `height` and `timestamp`. Both these fields are in `block_headers` table. 

But the query dispatches a number of other queries accessing multiple tables `transactions`, `block_deps`, `outputs`, `inputs` which are not needed other than just `block_headers`

Incase I'm overlooking something. Is this a valid issue? @tdroxler 

## Example un-tested query

Query to maybe use instead (un-tested)

```sql
SELECT (max(block_timestamp),
        sum(height),
        height)
FROM block_headers
WHERE chain_from = 1
  AND chain_to = 4
  AND height = (SELECT max(height)
                from block_headers
                WHERE chain_from = 1
                  AND chain_to = 4)
group by height;
```

## Use `UNION`

[The function](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L181) also traverses over multiple groups and queries them concurrently which can be replaced with `UNION`.

tdroxler:
daaaaamn you so right, that's the disadvantage to access data through the DAO only :/

Maybe we could give access to the DB directly or add new functions to the DAO.

That's a super nice catch :rocket:  I'll dig the code for more thing like that

tdroxler:
We also use `BlockDao.maxHeight` a lot, which could be replace using the `BlockCache`

tdroxler:
damn, [here](https://github.com/alephium/explorer-backend/blob/v1.8.0-rc1/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L280-L281) also we could use `getLite`.

The more and more I'm thinking of all this, the more I think we could be closer to the DB in the `BlockFlowSyncService`.
This is by far our heaviest sync service and we could maybe gain some perf to work directly with some `DBAction` rather than going trough `Future` in the BlockDao.

WDYT @simerplaha ?

simerplaha:
Yea totally! You are right. 

There are places here where we can squeeze out decent performance. 

[...]

PR #369, opened 2022-10-24
Replaced `BlockDao.get(parent)` with `BlockQueries.getBlockChainInfo(parent)`

simerplaha:
- Implements [this comment](https://github.com/alephium/explorer-backend/issues/328#issuecomment-1225295656) to reduce the amount of data read (inputs, outputs, transactions etc) when fetching `parent`.
- Replaced `assert(block.chainFrom == parentBlock.chainFrom && block.chainTo == parentBlock.chainTo)` with error `BlocksInDifferentChains` for better log message incase of error. 

It seems issue #358 is invalid after this because the full block is not being read anymore? Should it be closed?

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
@@ -74,6 +74,16 @@ object BlockQueries extends StrictLogging {
       header <- BlockHeaderSchema.table.filter(_.hash === hash).result.headOption
     } yield header.map(_.toLiteApi)
 
+  /** For a given `BlockHash` returns its basic chain information */
+  def getBlockChainInfo(hash: BlockHash): DBActionR[Option[(GroupIndex, GroupIndex, Boolean)]] =
+    BlockHeaderSchema.table
+      .filter(_.hash === hash)
+      .map { block =>
+        (block.chainFrom, block.chainTo, block.mainChain)
+      }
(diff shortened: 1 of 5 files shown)


app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
```diff
@@ -74,6 +74,16 @@ object BlockQueries extends StrictLogging {
...
 
+  /** For a given `BlockHash` returns its basic chain information */
+  def getBlockChainInfo(hash: BlockHash): DBActionR[Option[(GroupIndex, GroupIndex, Boolean)]] =
+    BlockHeaderSchema.table
+      .filter(_.hash === hash)
+      .map { block =>
+        (block.chainFrom, block.chainTo, block.mainChain)
+      }
+      .result
+      .headOption
```
tdroxler, line 85: should we try to directly use the raw sql format instead? I know it's a bit more painful, but we should  force our-self to do it (I'm the first one to blame for not doing it)

tdroxler (approved):
LGTM! one minor comment!

One more PR  that will improve syncing speed :rocket:

app/src/main/scala/org/alephium/explorer/persistence/queries/BlockQueries.scala
simerplaha: Ah yep. Done :)

Issue #344, opened 2022-09-09
Smart contract events support


polarker:
Details to be added.

@tdroxler Please feel free to reminder me if I forgot to discuss with you next week.

polarker:
Here is a draft proposal to enable to support:
1. Full node has an endpoint to query all contract events based on block hash. This index is disabled by default.
2. For each block hash, the explorer backend needs to sync both the block and the respective contract events. Explorer backend will persist events in the DB.
a. For each event, we create such indexes `(block hash, tx id, contract address, unique input address)`.
b. For simplicity, transactions with multiple input addresses will not be indexed.
3. Users could query for all of the events with parameters: `(contract address, user address)`

tdroxler:
When I sync a block, is there a way to know if this block has events or not? it could light the syncing process if we query the events only when necessary.

polarker:
> When I sync a block, is there a way to know if this block has events or not? it could light the syncing process if we query the events only when necessary.

There is no shortcut, but I could add a new endpoint `blockflow/blocks-with-events/{block-hash}` to fetch blocks and events in one shot. WDYT?

tdroxler:
That could be nice to light both the node and the explorer.

tdroxler:
when you say `input address` it's the address(es) in the `Val.Address` fields?

polarker:
> when you say `input address` it's the address(es) in the `Val.Address` fields?

By `input address`, I meant the address of TX inputs.

The address fields in the events are the contract addresses that emit the events. Which is another thing I will explain to you in more details later.

Issue #368, opened 2022-10-24
Centralise direct basic database queries in our test-cases to `TestQueries` object


simerplaha:
Updated:  3/11/2022

In our test-cases we directly access tables (`*Schema.table`) to access data.

For example:

```scala
run(BlockHeaderSchema.table ++= blockHeaders).futureValue
run(BlockHeaderSchema.table.delete).futureValue
run(BlockHeaderSchema.table.result).futureValue
run(BlockHeaderSchema.table.length.result).futureValue
```

Be nice to centralise even these basic table operations to the `TestQueries` object so we do not create a dependency on the target DB library we use everywhere in our test-cases.

tdroxler:
very good point, this would indeed ease a lot if one day we want to try another DB lib.

simerplaha:
Discussed this. And the current way is more convenient over creating dedicated functions in `TestQueries` which requires all `insert` and `delete` functions to be properly named because the compiler doesn't not allow overloading them which will be a pain in the a$$.

Preferring convenience of usage for now.

As the decision to use a SQL library (#355) progresses, this can be considered then.

PR #376, opened 2022-10-31
Feature/smart contract events

tdroxler:
First version of smart contract events, maybe we need more/others endpoints, but at least the data is in DB.

Resolves: https://github.com/alephium/explorer-backend/issues/344


app/src/main/scala/org/alephium/explorer/persistence/queries/EventQueries.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  def getEventsByContractAndInputAddressQuery(contract: Address,
+                                              input: Address): DBActionSR[EventEntity] =
+    sql"""
+      SELECT *
+      FROM events
+      WHERE contract_address = $contract
+      AND input_address = $input
+      ORDER BY block_timestamp DESC
+      """.asASE[EventEntity](eventGetResult)
```
simerplaha, line 87: Neat. Should there be indexes for these columns accessed here?
- `event_index`
- `block_timestamp`
- `input_address`

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
```diff
@@ -274,6 +274,17 @@ case object BlockFlowSyncService extends StrictLogging {
...
+      implicit ec: ExecutionContext,
+      dc: DatabaseConfig[PostgresProfile],
+      blockFlowClient: BlockFlowClient,
+      cache: BlockCache,
+      groupSetting: GroupSetting): Future[Unit] = {
+    for {
+      _ <- insert(blockWithEvents.block)
+      _ <- BlockDao.insertEvents(blockWithEvents.events)
+    } yield ()
+  }
```
simerplaha, line 287: If `insert` succeeds but `insertEvents` fails (or explorer gets stopped), then on retry/reboot, this not being a transaction wouldn't lead to an inconsistent database state?

app/src/test/scala/org/alephium/explorer/persistence/queries/EventQueriesSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+      }
+    }
+  }
+
+  trait Fixture {
+    def insert(events: ArraySeq[EventEntity]) = {
+      run(EventSchema.table.delete).futureValue
+      run(EventSchema.table ++= events).futureValue
+    }
+  }
```
simerplaha, line 164: Following issue #368 should these basic queries be moved to their dedicated `EventQueries` object?

This `Fixture` not being stateful means there's no need for the `Fixture` right? We could pull out `insert` to just a plain function sitting somewhere in the test?

app/src/test/scala/org/alephium/explorer/util/InputAddressUtilSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  trait Fixture {
+    def buildAssetInput(outputRef: api.model.OutputRef, unlockScript: protocol.vm.UnlockScript) = {
+      api.model.AssetInput(
+        outputRef,
+        serialize(unlockScript)
+      )
+    }
+  }
```
simerplaha, line 130: Same here. This `Fixture` being stateless means we can pull out `buildAssetInput` to just being a plain function sitting somewhere in tests?

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
```diff
@@ -0,0 +1,58 @@
...
+      .description("Get contract events by contract id")
+
+  val getEventsByContractAndInputAddress: BaseEndpoint[(Address, Address), ArraySeq[Event]] =
+    eventsEndpoint.get
+      .in("contract-address")
+      .in(path[Address]("contract_address"))
+      .in("input-address")
+      .in(path[Address]("input_address"))
+      .out(jsonBody[ArraySeq[Event]])
+      .description("Get contract events by contract id")
```
simerplaha, line 57: Should "Get contract events by contract id" be "Get contract events by contract id and input address"?

Just for my knowledge: Are "contract id" and "contract address" the same thing?

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
tdroxler: totally! I'll try to fix that

app/src/test/scala/org/alephium/explorer/persistence/queries/EventQueriesSpec.scala
tdroxler: oh indeed! yes let's try to follow #368 

you right with the Fixture, I have a bit of a reflex to use them, but it's not needed here.

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
tdroxler: nice catch, I'll fix it.

Yes contract id and contract address is the same, maybe I should rather use `contract-id` everywhere.

app/src/test/scala/org/alephium/explorer/persistence/queries/EventQueriesSpec.scala
tdroxler: oh still need to extract it out the `Fixture`

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
tdroxler: I think I'll rather use contract-address everywhere, as it's the notation received from the node https://github.com/alephium/alephium/blob/1e359e155b37c2afda6011cdc319d54ae8e4c059/api/src/main/scala/org/alephium/api/model/ContractEvents.scala#L43-L85

app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala
tdroxler: done

app/src/test/scala/org/alephium/explorer/util/InputAddressUtilSpec.scala
tdroxler: done

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
tdroxler: done

tdroxler:
@polarker I resurrected this PR from the abyss, ready to review.

simerplaha (approved):
🚀

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
```diff
@@ -0,0 +1,53 @@
...
+    def inputAddress: Rep[Option[Address]] = column[Option[Address]]("input_address")
+    def timestamp: Rep[TimeStamp]          = column[TimeStamp]("block_timestamp")
+    def eventIndex: Rep[Int]               = column[Int]("event_index")
+    def fields: Rep[Array[Byte]]           = column[Array[Byte]]("fields")
+
+    def * : ProvenShape[EventEntity] =
+      (blockHash, txHash, contractAddress, inputAddress, timestamp, eventIndex, fields)
+        .<>((EventEntity.apply _).tupled, EventEntity.unapply)
+
+    def pk: PrimaryKey = primaryKey("events_pk", (blockHash, txHash, contractAddress))
```
Lbqds, line 43: We cannot use `(blockHash, txHash, contractAddress)` as the primary key because a contract can emit multiple events in one transaction.

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
```diff
@@ -0,0 +1,58 @@
...
+      .description("Get contract events by contract address")
+
+  val getEventsByContractAndInputAddress: BaseEndpoint[(Address, Address), ArraySeq[Event]] =
+    eventsEndpoint.get
+      .in("contract-address")
+      .in(path[Address]("contract_address"))
+      .in("input-address")
+      .in(path[Address]("input_address"))
+      .out(jsonBody[ArraySeq[Event]])
+      .description("Get contract events by contract address")
```
h0ngcha0, line 57: [Nitpick] This has the same description as `getEventsByContractAddress`

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
```diff
@@ -0,0 +1,53 @@
...
+
+    def txHashIdx: Index          = index("tx_hash_idx", txHash)
+    def contractAddressIdx: Index = index("contract_address_idx", contractAddress)
+    def inputAddressIdx: Index    = index("input_address_idx", inputAddress)
+    def timestampIdx: Index       = index("block_timestamp_idx", timestamp)
+    def eventIndexIdx: Index      = index("event_index_idx", eventIndex)
```
h0ngcha0, line 49: Question: Is the `eventIndexIdx` index necessary? 
1. What is the use case?
2. It has a low cardinality, does it make sense performance wise if used individually?

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
```diff
@@ -0,0 +1,53 @@
...
+    def * : ProvenShape[EventEntity] =
+      (blockHash, txHash, contractAddress, inputAddress, timestamp, eventIndex, fields)
+        .<>((EventEntity.apply _).tupled, EventEntity.unapply)
+
+    def pk: PrimaryKey = primaryKey("events_pk", (blockHash, txHash, contractAddress))
+
+    def txHashIdx: Index          = index("tx_hash_idx", txHash)
+    def contractAddressIdx: Index = index("contract_address_idx", contractAddress)
+    def inputAddressIdx: Index    = index("input_address_idx", inputAddress)
+    def timestampIdx: Index       = index("block_timestamp_idx", timestamp)
```
h0ngcha0, line 48: Question: Do we need to query the timestamp? 
Maybe we need to query the block hash since we have an endpoint on full node to do that.

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -67,6 +70,16 @@ trait BlockFlowClient extends Service {
             .toArraySeq)
     }
 
+  def fetchBlocksAndEventsAtHeight(fromGroup: GroupIndex, toGroup: GroupIndex, height: Height)(
```
h0ngcha0, line 73: This method doesn't seem to be used any more.

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
tdroxler: oh should we add `eventIndex` in the `pk` then?

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
tdroxler: we don't query the timestamp, be we do a lot of `ORDER BY` with the timestamp.
I just read [this](https://www.postgresql.org/docs/current/indexes-ordering.html) and realize that most of our `ORDER BY` are using `DESC`, we could change our timestamp indexes with what's in the link above

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
tdroxler: We have [one query ](https://github.com/alephium/explorer-backend/blob/f1697856dc0a600d04b93583b1ddb4fdb3fb3e0f/app/src/main/scala/org/alephium/explorer/persistence/queries/EventQueries.scala#L65-L68) that order using the event_index, but it's maybe not worth as it as a low cardinality. WDYT?

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
Lbqds: No, a contract can emit multiple events with the same `eventIndex` in one transaction. Events emitted by a contract have an offset, but it's a bit tricky to get the event offset. And because of reorg, for the same event, different full nodes may have different event offsets. I'm not sure if it's a good idea to add the event offset to the primary key.

tdroxler: mmm then what would be the correct `PK`? maybe no primary key and use and auto increment? but we might have duplicate issue if we are not careful.

Lbqds: I think duplicate is no problem, events can also be duplicated on full node because of reorg, users need to check if the event is from the mainchain.

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
h0ngcha0: I see, thanks for the explanation <3

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
h0ngcha0: My feeling is that this query shouldn't be affected much since it is low cardinality :)

app/src/main/scala/org/alephium/explorer/persistence/queries/EventQueries.scala
```diff
@@ -0,0 +1,88 @@
...
+      WHERE tx_hash = $txId
+      ORDER BY event_index
+      """.asASE[EventEntity](eventGetResult)
+
+  def getEventsByContractAddressQuery(address: Address): DBActionSR[EventEntity] =
+    sql"""
+      SELECT *
+      FROM events
+      WHERE contract_address = $address
+      ORDER BY block_timestamp DESC
```
polarker, line 76: If two events are from the same block, we need to order them by the order too

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
```diff
@@ -0,0 +1,52 @@
...
+    def fields: Rep[Array[Byte]]           = column[Array[Byte]]("fields")
+
+    def * : ProvenShape[EventEntity] =
+      (blockHash, txHash, contractAddress, inputAddress, timestamp, eventIndex, fields)
+        .<>((EventEntity.apply _).tupled, EventEntity.unapply)
+
+    def pk: PrimaryKey = primaryKey("events_pk", (blockHash, txHash, contractAddress))
+
+    def txHashIdx: Index          = index("tx_hash_idx", txHash)
+    def contractAddressIdx: Index = index("contract_address_idx", contractAddress)
```
polarker, line 46: We may need a multi-index `(contract_address, timestamp, event_order)` if we want to paginate the events for addresses.

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
```diff
@@ -147,9 +161,12 @@ object BlockFlowClient extends StrictLogging {
 
     def fetchBlocks(fromTs: TimeStamp,
                     toTs: TimeStamp,
-                    uri: Uri): Future[ArraySeq[ArraySeq[BlockEntity]]] = {
-      _send(getBlocks, uri, api.model.TimeInterval(fromTs, toTs))
-        .map(_.blocks.map(_.map(blockProtocolToEntity).toArraySeq).toArraySeq)
+                    uri: Uri): Future[ArraySeq[ArraySeq[BlockEntityWithEvents]]] = {
```
polarker, line 164: should we deprecate `fetchBlocks` now?

app/src/main/scala/org/alephium/explorer/util/InputAddressUtil.scala
```diff
@@ -0,0 +1,72 @@
...
+   */
+  def addressFromProtocolInput(input: api.model.AssetInput): Option[Address] =
+    input.toProtocol() match {
+      case Right(value) =>
+        value.unlockScript match {
+          case protocol.vm.UnlockScript.P2PKH(pk) =>
+            Some(Address.unsafe(protocol.model.Address.p2pkh(pk).toBase58)): Option[Address]
+          case protocol.vm.UnlockScript.P2SH(script, _) =>
+            val lockup = protocol.vm.LockupScript.p2sh(protocol.Hash.hash(script))
+            Some(Address.unsafe(protocol.model.Address.from(lockup).toBase58)): Option[Address]
```
polarker, line 41: Type annotations in the above two case branches can be removed.

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
polarker: We could introduce a new field `event_order_in_block` for the index of an event inside a block, like how we handle tx orders inside a block. Then we could use `(block_hash, event_order_in_block)` as the primary key.

app/src/main/scala/org/alephium/explorer/api/model/Event.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.alephium.explorer.api.Codecs._
+import org.alephium.json.Json._
+import org.alephium.protocol.model.{BlockHash, TransactionId}
+
+final case class Event(
+    blockHash: BlockHash,
+    txHash: TransactionId,
+    contractAddress: Address,
+    inputAddress: Option[Address],
+    eventIndex: Int,
```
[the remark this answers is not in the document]
tdroxler: I put it as it's also in the [node api model](https://github.com/alephium/alephium/blob/master/api/src/main/scala/org/alephium/api/model/ContractEvents.scala#L39), we need them in the db to sort, but we can remove them from the api if you don't think it will be useful for anyone.

app/src/main/scala/org/alephium/explorer/util/InputAddressUtil.scala
```diff
@@ -0,0 +1,72 @@
...
+      val addresses = inputs
+        .map(InputAddressUtil.addressFromProtocolInput)
+        .collect { case Some(address) => address }
+
+      if (addresses.sizeIs == inputs.size && addresses.distinct.sizeIs == 1) {
+        Some(addresses.head)
+      } else {
+        None
+      }
+    }
```
polarker, line 70: Let's use this optimized version:
```
    val addressOpt = inputs.headOption.flatMap(InputAddressUtil.addressFromProtocolInput)
    addressOpt match {
      case None => None
      case Some(_) =>
        if (inputs.tail.forall(input => InputAddressUtil.addressFromProtocolInput(input) == addressOpt)) {
          addressOpt
        } else {
          None
        }
    }
```

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
```diff
@@ -0,0 +1,58 @@
...
+      .in(path[TransactionId]("transaction_id"))
+      .out(jsonBody[ArraySeq[Event]])
+      .description("Get contract events by transaction id")
+
+  val getEventsByContractAddress: BaseEndpoint[Address, ArraySeq[Event]] =
+    eventsEndpoint.get
+      .in("contract-address")
+      .in(path[Address]("contract_address"))
+      .out(jsonBody[ArraySeq[Event]])
+      .description("Get contract events by contract address")
```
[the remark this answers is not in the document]
tdroxler: and why not doing pagination?

app/src/main/scala/org/alephium/explorer/api/model/Event.scala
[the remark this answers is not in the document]
polarker: Sorry, this field is very useful. Forgive my comments 😅

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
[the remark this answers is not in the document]
polarker: dApps mostly only care about latest events and it might be costy to paginate all of the events.

But I agree we could start with pagination and see.

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
tdroxler: yes we could, nothing urgent as it's for internal use only.

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
tdroxler: It makes totally sense, I misunderstood `event_index` and treated it as `even_order_in_block`

app/src/main/scala/org/alephium/explorer/service/BlockFlowClient.scala
tdroxler: it's actually still used in our `SanityChecker`

app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala
tdroxler: I did try to add the index, but the `EXPLAIN` is the same with or without the index, but as we saw in the past the output of the explain can be different depending of the amount of data. I currently don't have any event so it's maybe not relevant.

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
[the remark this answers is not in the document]
tdroxler: pagination done

app/src/main/scala/org/alephium/explorer/api/EventsEndpoints.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  val getEventsByContractAndInputAddress: BaseEndpoint[(Address, Address), ArraySeq[Event]] =
```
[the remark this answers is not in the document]
tdroxler: pagination done

polarker (approved):
LGTM 🚀

Issue #323, opened 2022-08-16
`org.postgresql.util.PSQLException`: ERROR: deadlock detected


simerplaha:
`ERROR: deadlock detected` in [build](https://github.com/alephium/explorer-backend/runs/7850144431?check_suite_focus=true#step:5:2480) - `TokenSupplyServiceSpec`

> [info] TokenSupplyServiceSpec:
[info] org.alephium.explorer.service.TokenSupplyServiceSpec *** ABORTED ***
[info]   org.postgresql.util.PSQLException: ERROR: deadlock detected
[info]   Detail: Process 1425 waits for AccessExclusiveLock on relation 16750 of database 13757; blocked by process 1426.
[info] Process 1426 waits for RowExclusiveLock on relation 16728 of database 13757; blocked by process 1425.
[info]   Hint: See server log for query details.
[info]   at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2533)
[info]   at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2268)
[info]   at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:313)
[info]   at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:448)
[info]   at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:369)
[info]   at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:159)
[info]   at org.postgresql.jdbc.PgPreparedStatement.execute(PgPreparedStatement.java:148)
[info]   at slick.jdbc.StatementInvoker.results(StatementInvoker.scala:39)
[info]   at slick.jdbc.StatementInvoker.iteratorTo(StatementInvoker.scala:22)
[info]   at slick.jdbc.Invoker.first(Invoker.scala:30)
[info]   ...

simerplaha:
## Possible cause

The good old deadlock issue. From the error, I'm assuming that it occurred because `sbt` runs each test class concurrently so we got witness this rare event. The error message mentions two processes `1425` and `1426` so I'm assuming two tests run by `sbt` in two different threads. 

## Replicating issue locally

Looking at the queries executed by `TokenSupplyServiceSpec` to figure out where this could happen -`BlockQueries.updateMainChainStatusAction` seems to be the most likely one to cause this.

[...]

PR #377, opened 2022-10-31
Improve test framework

tdroxler:
Big refactor for our test framework:

- Define `AlephiumFutureSpec` to clean the usage of `ScalaFutures/PatienceConfig` which was spread all over the place
- Define an `ExecutionContext` inside `AlephiumFutureSpec` to avoid declaring it everywhere.

Finally the big change that started this PR, which I think will fix https://github.com/alephium/explorer-backend/issues/323

Until now we were using the default `postgres` database for all our test and we were dropping/recreating the tables for each tests.
With this PR we are now creating one database per test, we thus avoid conflict where two parallel test are using the same databae (which lead to the deadlocks).

I did re-run all jobs several time without any issue so far :crossed_fingers:


simerplaha (approved):
Nice! 🚀 🚀 🚀

Issue #375, opened 2022-10-28
Endpoint to get address' transasactions by time range


tdroxler:
This could be useful for the front end to implement this https://github.com/alephium/explorer/issues/41
We need to be careful when choosing the max time range, a big one could be costly for addresses with lots of transactions

we already have the `/addresses/<address>/transactions` , we  could have `/addresses/<address>/timeranged-transactions?fromTs=x&toTs=x`

tdroxler:
probably we won't be able to use a full year, as it could take too long to return a query for the big addresses.
Is it fine to loop on your side to get all the data you want?

mvaivre:
I fear that this might get a bit messy... 
Also, the "full year" is probably the most interesting option for task purposes.

Have you made some benchmark tests, to check how long it takes to extract a full year worth of TXs for an address?

tdroxler:
>Have you made some benchmark tests, to check how long it takes to extract a full year worth of TXs for an address?

Not yet, but will do as soon as the feature is there!

simerplaha:
To get a sense of how long it takes to extract "a full years worth of TXs", here are some benchmarks of the query specific to PR. 

We get 
- 13.2 queries/second for reading an address with 50,000 transactions
- 1.1 queries/second for reading an address with 500,000 transactions
- Generated data
  - Address count: 1
  - A unique `block_timestamp` was generated per 5 transactions (so 5 transactions per address per block).
      - So 10,000 unique `block_timestamp`s for first benchmark
      -  And 100,000 unique `block_timestamp`s for second benchmark
  - With HikariCP disabled 
 
Obviously as more addresses get added to the table the performance will decrease. For example: With 100 address the above performance of 13.2 queries/second reduces to 9 queries/second.

I might be wrong here but "a full years worth of TXs" is a bit subjective right? Because in theory an address could have millions transactions over a period of month or only 1 transaction over a period of couple of years. So adding restriction on time period range might not address the concern of fetching too much data?

Maybe we want to limit the maximum number of transactions returning instead of time-period using `Pagination`? 

For example: If we want to limit to a max of `100,000` transactions per request, UI could dispatch a request for `100,000 + 1` with the requested time-period range. If the response has `100,000 + 1` transactions, another request could be dispatched to fetching the remaining using pagination and if the response contains less than `100,000` transactions that completes the request?

#### Benchmark screenshots

[...]

PR #378, opened 2022-11-02
#375: Time ranged transactions

simerplaha:
- First version for #375
- Pagination was required so backend does not allow large limits to go through. `100000` being the max limit. Defaults to `20` just like other APIs (Uses existing pagination code). 
- Added `TestQueries` for issue #368. Didn't seem like it was nice to add queries like `clearTable()` to production code. Will update that issue's description if you ok with this.
- Is there a tool that generates `explorer-backend-openapi.json`? Updated this one here by hand.

`/addresses/[address]/timeranged-transactions?fromTs=x&toTs=y`
OR
`/addresses/[address]/timeranged-transactions?fromTs=x&toTs=y&limit=100000`
OR 
`/addresses/[address]/timeranged-transactions?fromTs=x&toTs=y&limit=100000&page=1`

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
@@ -23,6 +23,7 @@ object Pagination {
   val defaultPage: Int  = 1
   val defaultLimit: Int = 20
   val maxLimit: Int     = 100
+  val `100K`: Int     = 100_000
 
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   def unsafe(offset: Int, limit: Int, reverse: Boolean = false): Pagination = {
--- a/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
@@ -23,7 +23,7 @@ object Pagination {
   val defaultPage: Int  = 1
   val defaultLimit: Int = 20
   val maxLimit: Int     = 100
-  val `100K`: Int     = 100_000
+  val `100K`: Int       = 100000
 
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   def unsafe(offset: Int, limit: Int, reverse: Boolean = false): Pagination = {
--- a/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
@@ -23,7 +23,7 @@ object Pagination {
   val defaultPage: Int  = 1
   val defaultLimit: Int = 20
   val maxLimit: Int     = 100
-  val `100K`: Int       = 100000
+  val thousand: Int     = 1000
 
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   def unsafe(offset: Int, limit: Int, reverse: Boolean = false): Pagination = {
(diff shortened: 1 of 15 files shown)


tdroxler:
>Is there a tool that generates explorer-backend-openapi.json? Updated this one here by hand.

yep: `sbt tools/run`

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -178,6 +178,33 @@ object TransactionQueries extends StrictLogging {
...
+                                                   fromTime: TimeStamp,
+                                                   toTime: TimeStamp,
+                                                   offset: Int,
+                                                   limit: Int): DBActionSR[TxByAddressQR] = {
+    sql"""
+      SELECT tx_hash, block_hash, block_timestamp, tx_order
+      FROM transaction_per_addresses
+      WHERE main_chain = true
+        AND address = $address
+        AND block_timestamp between $fromTime AND $toTime
```
tdroxler, line 200: niiice didn't know about `BETWEEN`, let's maybe capitalize it so we know it's a keyword

app/src/test/scala/org/alephium/explorer/TestQueries.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package org.alephium.explorer
+
+import slick.jdbc.PostgresProfile.api._
+
+import org.alephium.explorer.persistence._
+import org.alephium.explorer.persistence.model.TransactionPerAddressEntity
+import org.alephium.explorer.persistence.schema.TransactionPerAddressSchema
+
+object TestQueries {
```
tdroxler, line 25: yep better to have such queries here rather than in production code

app/src/test/scala/org/alephium/explorer/persistence/queries/TransactionQueriesSpec.scala
```diff
@@ -439,6 +439,118 @@ class TransactionQueriesSpec
...
+
+                  //only the main_chain entities within the time-range is expected
+                  actualResult should contain theSameElementsAs expectedResult
+              }
+          }
+        }
+      }
+    }
+  }
+
```
tdroxler, line 553: you are a true scalacheck expert :rocket:

app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
```diff
@@ -23,6 +23,7 @@ object Pagination {
   val defaultPage: Int  = 1
   val defaultLimit: Int = 20
   val maxLimit: Int     = 100
+  val `100K`: Int     = 100_000
```
tdroxler: oh cool didn't know we can use `_` in `Int`, this makes it much more readable

app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
```diff
@@ -23,7 +23,7 @@ object Pagination {
   val defaultPage: Int  = 1
   val defaultLimit: Int = 20
   val maxLimit: Int     = 100
-  val `100K`: Int     = 100_000
+  val `100K`: Int       = 100000
```
tdroxler, line 26: ah, any reason you went back to no `_`?

tdroxler (commented):
LGTM, but I'll just test locally before aproving

tdroxler:
ok so good news the endpoints works well, but the bad news: not with 100K, it's a bit too heavy with the number of data we have in our prod DB. I talk with Mika we will start with 1000 and start experimenting wiith this.

Can you change to 1000? I'll approve, so you can merge once you change to that.

Thx :rocket:

tdroxler (approved):

app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala
simerplaha: scalafmt didn't like it: `error: Invalid literal number`

tdroxler: oh ok, I'll check if we can change that, I kinda like this format

PR #380, opened 2022-11-08
#337: Implements `WriteOnly`/`SyncOnly` mode

### LINKED ISSUE #337: Introduce a `SyncOnly` application
Split the code so we can have two applications: 

One to do synchronization only and  one ReadOnly that exposes the endpoints.

We still want to be able to launch a `ReadWriteApplication` as today.

simerplaha:
- Resolves #337
- Added `WriteOnly` (or `SyncOnly` mode). Named it `WriteOnly` to follow the existing naming convention. 
- Uses `ExecutionContext.Implicits.global` for `WriteOnly` mode since `ActorSystem` is not available.

## Configuration

```properties
# ReadOnly = Starts Explorer in read-only mode
# WriteOnly = Starts Explorer in write-only mode
# ReadWrite = Starts Explorer in read-write mode
boot-mode = WriteOnly
```

### DIFF
--- a/app/src/main/resources/application.conf
+++ b/app/src/main/resources/application.conf
@@ -12,7 +12,7 @@ explorer {
     # WriteOnly = Starts Explorer in write-only mode
     # ReadWrite = Starts Explorer in read-write mode
     boot-mode = ReadWrite
-    boot-mode = ${?EXPLORER_READONLY}
+    boot-mode = ${?EXPLORER_BOOT_MODE}
 
     # Sync interval for BlockFlowSyncService & MempoolSyncService
     sync-period = 5 seconds
--- a/app/src/main/resources/application.conf
+++ b/app/src/main/resources/application.conf
@@ -8,10 +8,11 @@ explorer {
     host = "localhost"
     host = ${?EXPLORER_HOST}
 
-    # true = Starts Explorer in read-only mode
-    # false = Starts Explorer in read-write mode
-    read-only = false
-    read-only = ${?EXPLORER_READONLY}
+    # ReadOnly = Starts Explorer in read-only mode
+    # WriteOnly = Starts Explorer in write-only mode
+    # ReadWrite = Starts Explorer in read-write mode
+    boot-mode = ReadWrite
+    boot-mode = ${?EXPLORER_READONLY}
 
     # Sync interval for BlockFlowSyncService & MempoolSyncService
     sync-period = 5 seconds
--- a/app/src/main/scala/org/alephium/explorer/Main.scala
+++ b/app/src/main/scala/org/alephium/explorer/Main.scala
@@ -36,7 +36,7 @@ object Main extends StrictLogging {
       (new BootUp).init()
     } catch {
       case error: Throwable =>
-        logger.error(s"Cannot initialize system: $error")
+        logger.error("Cannot initialize system", error)
     }
   }
 }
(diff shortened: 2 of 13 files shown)


app/src/main/scala/org/alephium/explorer/Main.scala
```diff
@@ -55,32 +55,70 @@ class BootUp extends StrictLogging {
...
     Runtime.getRuntime.addShutdownHook(new Thread(() => {
       sideEffect(Await.result(actorSystem.terminate(), 10.seconds))
     }))
+
+    CoordinatedShutdown(actorSystem).addTask(
+      CoordinatedShutdown.PhaseBeforeActorSystemTerminate,
+      "Shutdown services"
+    ) { () =>
+      for {
+        _ <- explorer.stop()
```
tdroxler, line 95: :+1: That might solve the issue raised by @killerwhile in slack

app/src/main/scala/org/alephium/explorer/Main.scala
```diff
@@ -55,32 +55,70 @@ class BootUp extends StrictLogging {
...
 
-  CoordinatedShutdown(actorSystem).addTask(
-    CoordinatedShutdown.PhaseBeforeActorSystemTerminate,
-    "Shutdown services"
-  ) { () =>
-    for {
-      _ <- explorer.stop()
-    } yield Done
+  @SuppressWarnings(Array("org.wartremover.warts.GlobalExecutionContext"))
+  def initInWriteOnlyMode(): Unit = {
```
tdroxler, line 101: I was going to ask the difference with the other, but no Actor-system!!! sweet!

app/src/main/scala/org/alephium/explorer/Main.scala
```diff
@@ -55,32 +55,70 @@ class BootUp extends StrictLogging {
...
+    implicit val ec: ExecutionContext =
+      scala.concurrent.ExecutionContext.Implicits.global
```
tdroxler, line 103: we could create a task about properly configure an Execution context, I think there are way to avoid using the global one and be more fined grained about what we want, but I almost always used Akka, so I don't remember how ^^

tdroxler (approved):
Super clean! I love it :rocket: I'll test it right away

tdroxler:
I successfully launched a `WriteOnly` and `ReadOnly` in parallel :rocket:

app/src/main/resources/application.conf
```diff
@@ -8,10 +8,11 @@ explorer {
...
 
-    # true = Starts Explorer in read-only mode
-    # false = Starts Explorer in read-write mode
-    read-only = false
-    read-only = ${?EXPLORER_READONLY}
+    # ReadOnly = Starts Explorer in read-only mode
+    # WriteOnly = Starts Explorer in write-only mode
+    # ReadWrite = Starts Explorer in read-write mode
+    boot-mode = ReadWrite
+    boot-mode = ${?EXPLORER_READONLY}
```
tdroxler, line 15: FYI @killerwhile here is the new configuration! We will very soon (probably today) release a new version. I think we'll directly include this.

Oh and @simerplaha you need to change `EXPLORER_READONLY` to `EXPLORER_BOOT_MODE` otherwise it will receive a boolean

app/src/main/scala/org/alephium/explorer/Main.scala
simerplaha: Oh this code is exactly the same as before. No new changes here. Just moved it to a function.

app/src/main/resources/application.conf
simerplaha: Whoops! Good catch. Done.

app/src/main/scala/org/alephium/explorer/Main.scala
tdroxler: ah ok ^^

app/src/main/resources/application.conf
tdroxler: great, good to merge then!

app/src/main/scala/org/alephium/explorer/Main.scala
simerplaha: I think we've already got one #204. Can look at tuning stuff there but I'm beginning to realise that there might be no need for special tuning other than just in `ReadWrite` mode. But yep this can be looked into within that issue?

PR #387, opened 2022-11-16
#299: Endpoint to fetch transactions for a list of addresses

### LINKED ISSUE #299: Provide endpoint to fetch list of transactions for a list of addresses
In the desktop wallet the pagination of the transaction list of the overview page is problematic.

When first rendering the Overview page, we fetch the transactions page 1 of each address of the wallet, then we merge this list, and we sort it by date. When clicking "Load more" we fetch the transactions page 2 of each address, merge the data with the previous one and sort them again. This might cause the following issue: A transaction of address X of page 2 might have an earlier date than a transaction of address Y of page 1, causing a reshuffling of the transaction list, which is bad UX.

I would like to ask if it's possible to have an endpoint on backend that returns the list of transactions for a set of addresses so that we can easily paginate it. Same way we have an endpoint to fetch a (sorted by date) paginated list of transactions for an address, I'd like the same for a list of addresses.

simerplaha:
- Resolves #299
- Following `areAddressesActive` this endpoint is also a `POST` request. 
  - Endpoint: `/addresses/transactions`
  - Requires an array of addresses as JSON body `["address1", "address2"]`

### DIFF
--- a/app/src/test/scala/org/alephium/explorer/ExplorerSpec.scala
+++ b/app/src/test/scala/org/alephium/explorer/ExplorerSpec.scala
@@ -293,6 +293,25 @@ trait ExplorerSpec
     }
   }
 
+  "get all transactions for addresses" in {
+    forAll(Gen.someOf(transactions)) { transactions =>
+      val limitedAddresses = transactions.flatMap(_.outputs.map(_.address)).take(txLimit)
+      val addressesBody    = limitedAddresses.map(address => s""""$address"""").mkString("[", ",", "]")
+
+      Post("/addresses/transactions", addressesBody) check { response =>
+        val expectedTransactions =
+          transactions.filter(_.outputs.exists(limitedAddresses.contains)).take(txLimit)
+
+        val res = response.as[ArraySeq[Transaction]]
+
+        res.size is limitedAddresses.size
+        Inspectors.forAll(expectedTransactions) { transaction =>
+          res.map(_.hash) should contain(transaction.hash)
+        }
+      }
+    }
+  }
+
   "generate the documentation" in {
     Get("/docs") check { response =>
       response.code is StatusCode.Ok
--- a/app/src/test/scala/org/alephium/explorer/web/AddressServerSpec.scala
+++ b/app/src/test/scala/org/alephium/explorer/web/AddressServerSpec.scala
@@ -128,22 +128,7 @@ class AddressServerSpec()
   }
 
   "respect the max number of addresses" in {
-    forAll(addressGen) {
-      case (address) =>
-        val size = groupSetting.groupNum * 20
-
-        val jsonOk = s"[${ArraySeq.fill(size)(s""""$address"""").mkString(",")}]"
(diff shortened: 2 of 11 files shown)


tdroxler:
pinging @nop33 just to validate it's what you need

tdroxler (changes requested):
Can you add a test in `ExplorerSpec` please? This file is a bit like our it:test, it makes sure every part of the code is well plugged when calling the endpoint.

Otherwise LGTM

nop33:
What is the limit of number of addresses that I can provide in the JSON body?

simerplaha:
> What is the limit of number of addresses that I can provide in the JSON body?

Max limit is `80` addresses (`groupNum` * [`gapLimit`](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit) i.e. `4 * 20`) similar to `/addresses-active`

Over `80` and the response is 

```json
{
    "detail": "Invalid value for: body (expected size of value to be less than or equal to 80, but was 81)"
}
```

> Can you add a test in ExplorerSpec please? This file is a bit like our it:test, it makes sure every part of the code is well plugged when calling the endpoint.

Added tests to `ExplorerSpec` and `AddressServerSpec`. Please let me know if this is what you expected.

Adding another test to check "txLimit" which would mean abstracting and reusing [this](https://github.com/alephium/explorer-backend/blob/cfc90f0d8ce9ff67fd5a59c107075b3cf1840d91/app/src/test/scala/org/alephium/explorer/web/AddressServerSpec.scala#L68) test. Don't wanna go too overboard but let me know if you'd like this.

app/src/test/scala/org/alephium/explorer/ExplorerSpec.scala
```diff
@@ -293,6 +293,25 @@ trait ExplorerSpec
     }
   }
 
+  "get all transactions for addresses" in {
+    forAll(Gen.someOf(transactions)) { transactions =>
+      val limitedAddresses = transactions.flatMap(_.outputs.map(_.address)).take(txLimit)
+      val addressesBody    = limitedAddresses.map(address => s""""$address"""").mkString("[", ",", "]")
```
tdroxler, line 299: it's a shame, but I actually discovered yesterday that you can pass 3 args to `mkString` and not only 1 :sweat_smile:

app/src/test/scala/org/alephium/explorer/web/AddressServerSpec.scala
```diff
@@ -154,4 +139,34 @@ class AddressServerSpec()
...
+        }
+      }
+    }
+
+    "respect the max number of addresses" in {
+      forAll(addressGen)(respectMaxNumberOfAddresses("/addresses/transactions", _))
+    }
+  }
+
+  def respectMaxNumberOfAddresses(endpoint: String, address: Address) = {
```
tdroxler, line 157: great idea :+1: it's good to add such test for each endpoint having an address limit. I see it as a "regression test", we make sure we don't remove the limit by mistake.

tdroxler (approved):
LGTM, thx!

Issue #386, opened 2022-11-14
`BlockDaoSpec` is super slow


tdroxler:
I haven't investigate more, it might be due to the tests using the caches.

PR #390, opened 2022-11-18
#386: Configure low cache timeout for test-cases

simerplaha:
- First PR for #386
- As you mentioned in the issue, the cache is the problem. Cache timeout being 10.second caused delayed test completion. This reduces cache timeout to `1.second` for test-cases. 
- Rough benchmarks
  - Test that took 1 minute 21 seconds now finishes in 22 seconds
  - Test that took 3 minutes 2 seconds now finishes in 43 seconds
     - Out of the 43 seconds, roughly 30 second are used by data generation `genBlockEntityWithOptionalParent()`
     
If you are ok with this change, will replicate it for other caches and test-cases accessing `BlockCache`.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
+++ b/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
@@ -52,9 +52,11 @@ object BlockCache {
   }
 
   // scalastyle:off
-  def apply()(implicit groupSetting: GroupSetting,
-              ec: ExecutionContext,
-              dc: DatabaseConfig[PostgresProfile]): BlockCache = {
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+  def apply(cacheRowCountReloadTime: FiniteDuration = 10.seconds)(
+      implicit groupSetting: GroupSetting,
+      ec: ExecutionContext,
+      dc: DatabaseConfig[PostgresProfile]): BlockCache = {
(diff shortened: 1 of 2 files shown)


app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
```diff
@@ -52,9 +52,11 @@ object BlockCache {
   }
 
   // scalastyle:off
-  def apply()(implicit groupSetting: GroupSetting,
-              ec: ExecutionContext,
-              dc: DatabaseConfig[PostgresProfile]): BlockCache = {
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+  def apply(cacheRowCountReloadTime: FiniteDuration = 10.seconds)(
```
tdroxler, line 56: what about removing the default argument and add our reload times in our configuration?
If you prefer you can create a new task for this.

tdroxler:
>Out of the 43 seconds, roughly 30 second are used by data generation genBlockEntityWithOptionalParent() 

I was going to suggest to use `Gen.listOfN(x, ...)` but I checked and the size of the list with `Gen.listOf(...)` vary between 0 and around 50-60, so it's all good, I was worried it creates 1000 elements.

So  let's stay like this for now.

tdroxler (approved):
LGTM for this one, for the next reload time, let's decide if we move them to our configuration first.

PR #393, opened 2022-11-22
Fix `java.util.NoSuchElementException: None.get`

simerplaha:
Calls to `BlockCache.getAllLatestBlocks` throw the following `NoSuchElementException` when the cache is empty or partially populated.

> java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:627)
	at scala.None$.get(Option.scala:626)
	at org.alephium.explorer.cache.BlockCache$.$anonfun$apply$2(BlockCache.scala:69)

This PR removes [this](https://github.com/alephium/explorer-backend/blob/09b30e9e7be13dd737da6036ac67fd56b7ee871e/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala#L69) call to `Option.get`.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
+++ b/app/src/main/scala/org/alephium/explorer/cache/BlockCache.scala
@@ -60,6 +60,14 @@ object BlockCache {
       dc: DatabaseConfig[PostgresProfile]): BlockCache = {
     val groupConfig: GroupConfig = groupSetting.groupConfig
 
+    /*
+     * `Option.get` is used to avoid unnecessary memory allocations.
+     * This cache is guaranteed to have data available for
+     * all chain-indexes after a single run of sync so `.get` would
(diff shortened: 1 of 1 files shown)


simerplaha:
I think I might not have consider the tradeoff here. Getting rid of that `NoSuchElementException` with and `Option` is nice but it also means we pay for allocation every time `BlockCache.getAllLatestBlocks` is invoked.

Since sync is always going to populate at least one row for each `ChainIndex` in production, this exception would never occur in production. 

Should this be closed?

tdroxler:
Yes I think we are safe to keep `_.get` here, maybe we could rather add a very small comment why we have it?

simerplaha:
Yep totally 👍🏼

simerplaha:
Ah I forgot about this issue. 

Dropped the commit and added a small comment.

tdroxler (approved):

Issue #388, opened 2022-11-17
Endpoints can be declared as `def`


simerplaha:
Currently all endpoints are either `val` or `lazy val`. For example [AddressesEndpoints](https://github.com/alephium/explorer-backend/blob/cfc90f0d8ce9ff67fd5a59c107075b3cf1840d91/app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala#L51).

There isn't a need to keep endpoint references alive throughout the life of the application and can be discarded after the router registers them. These endpoints can be declared as `def`.

Is this a valid issue?

tdroxler:
this seems totally legit, all the examples given by `tapir` are using `val` but it indeed doesn't look mandatory.

we are actually using twice the endpoint, in the Servers and the Documentation, but anyway since the explorer-backend the run for several weeks, beter to discard them.

 Don't hesitate to tackle this whenever you want, it's small fast one.

tdroxler:
@polarker let's continue the technical discussion here.

The question here is: Do we want to have our endpoints defined as `def`.

All our endpoints are called only twice, in their `Server` and in `Documentation`, once the routes are initialized, the reference is kept in the `routes` val. [here](https://github.com/alephium/explorer-backend/blob/b4a9bb0b04a16398c49f2bf0a596b0a2e22a0d77/app/src/main/scala/org/alephium/explorer/web/DocumentationServer.scala#L31-L35) and [here](https://github.com/alephium/explorer-backend/blob/b4a9bb0b04a16398c49f2bf0a596b0a2e22a0d77/app/src/main/scala/org/alephium/explorer/ExploreHttpServer.scala#L33)

If you use `val`, each endpoint will be initialized only once on start, but we will keep a reference for each of them without using them anymore.

If we used `def`, each endpoint will be initialized twice on start, but then we don't need to keep the allocation.

I agree the allocation isn't big at all, so we won't see a difference, but if we let the explorer-backend run for months without restarting, it sounds "sad" (don't know how to express it) to keep everything while we know it won't be used anymore.

Like keeping shoes that are too small in your closet ^^

WDYT?

PR #394, opened 2022-11-23
#388: `BlockEndpoints` and `ChartEndpoints` endpoints as `def`

simerplaha:
- First PR for #388
- Replaces Endpoint classes to be `object`s instead of `trait`s
- Sets endpoints to be defined as functions instead of `val`s

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/api/BlockEndpoints.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/BlockEndpoints.scala
@@ -22,33 +22,32 @@ import sttp.tapir._
 import sttp.tapir.generic.auto._
 
 import org.alephium.api.{alphJsonBody => jsonBody}
-import org.alephium.explorer.api.BaseEndpoint
 import org.alephium.explorer.api.model._
 import org.alephium.protocol.model.BlockHash
 
-trait BlockEndpoints extends BaseEndpoint with QueryParams {
+object BlockEndpoints extends BaseEndpoint with QueryParams {
 
-  private val blocksEndpoint =
+  private def blocksEndpoint() =
     baseEndpoint
       .tag("Blocks")
       .in("blocks")
 
-  val getBlockByHash: BaseEndpoint[BlockHash, BlockEntryLite] =
-    blocksEndpoint.get
+  def getBlockByHash(): BaseEndpoint[BlockHash, BlockEntryLite] =
+    blocksEndpoint().get
       .in(path[BlockHash]("block_hash"))
       .out(jsonBody[BlockEntryLite])
       .description("Get a block with hash")
 
-  val getBlockTransactions: BaseEndpoint[(BlockHash, Pagination), ArraySeq[Transaction]] =
-    blocksEndpoint.get
+  def getBlockTransactions(): BaseEndpoint[(BlockHash, Pagination), ArraySeq[Transaction]] =
+    blocksEndpoint().get
       .in(path[BlockHash]("block_hash"))
       .in("transactions")
       .in(pagination)
       .out(jsonBody[ArraySeq[Transaction]])
       .description("Get block's transactions")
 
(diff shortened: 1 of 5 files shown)


app/src/main/scala/org/alephium/explorer/api/BlockEndpoints.scala
```diff
@@ -22,33 +22,32 @@ import sttp.tapir._
...
 import org.alephium.api.{alphJsonBody => jsonBody}
-import org.alephium.explorer.api.BaseEndpoint
 import org.alephium.explorer.api.model._
 import org.alephium.protocol.model.BlockHash
 
-trait BlockEndpoints extends BaseEndpoint with QueryParams {
+object BlockEndpoints extends BaseEndpoint with QueryParams {
 
-  private val blocksEndpoint =
+  private def blocksEndpoint() =
```
tdroxler, line 30: if there are no differences, what do you think of defining:

```
private def blocksEndpoint
```
without the `()`? we wouldn't need to add them everywhere, as we wont' have args anyway.

This makes it a bit nicer IMO, but maybe I miss something?

simerplaha: Just a preference. Parens `()` shows at the call site that what's being called is a function. If this funtion does need parameters in the future then changing it makes the call site a curried function which in some odd cases would skips compile check. 

It's just a preference thing. I use parens `()` in my projects mostly because it's self documenting. 

No biggie. Parens removed.

tdroxler (approved):

app/src/main/scala/org/alephium/explorer/api/BlockEndpoints.scala
polarker: `omission of parentheses` can be dangerous as the function might has side effects.

When `val` is possible, I believe that `(lazy) val` better than `def blocksEndpoint()` better than `def blocksEndpoint`. With `val` I am sure that it will be initialized only once, while `def` cannot.

The cost is just a few object references (4 bytes each). The extra cost is pretty low. I would consider removing those references as over-optimizations.

polarker: It's as simple as this:
1. I provide my feedback.
2. If the feedback makes sense, then let's take it. If it does not make sense, just comment back and let me know.

It's never `whatever you want`.

polarker: I think we both agree that `do whatever you want` and `do whatever I want` are both too extreme and bad for the project.

simerplaha: The choice of words in your feedback is not starting a discussion. It’s telling to do something you think is right.

You have strong opinions and I really have no desire to counter your decisions even if it doesn't make sense. It's just wasted time.

This PR is not `do whatever I want`. Like every other PR an issue gets created to confirm its validity or discussed on Slack with Thomas and only then a PR gets submitted.

Please let me know how you’d like this PR progressed because I don’t know.

polarker: It's never `whatever you want`. -> It's never `whatever I want`.

I copied that phrase from your sentence and forgot to replace `you` with `I`.

polarker: Re: how to progress

I only looked at the discussion in this thread, I did not look at the other code of this PR. It's not me to decide how this PR should move forward as I don't review it at all.

> You have strong opinions and I really have no desire to counter your decisions even if it doesn't make sense. It's just wasted time.

I don't know how this kind of collaboration could work for the project for the long term, but let's discuss this off Github.

tdroxler:
Let's close those PRs until the discussion [here](https://github.com/alephium/explorer-backend/issues/388) is done,  we can re-open later if needed.

PR #397, opened 2022-11-24
#388: `AddressesEndpoints` as `def`

simerplaha:
- Third PR for #388
- This PR is stacked on #396
- Replaces `AddressesEndpoints` & `Documentation` to be `object`s instead of `trait`s
- Sets endpoints to be defined as functions instead of `val`s

FYI: I've manually checked the APIs I'm familiar with to make sure they are running as usual. Hoping the tests will ensure that all the rest of the APIs are ok (too many to check manually anyway).


tdroxler (approved):

tdroxler:
see https://github.com/alephium/explorer-backend/pull/394#issuecomment-1331756069

Issue #155, opened 2022-02-28
Migrate main chain update to raw sql


tdroxler:
[this code](https://github.com/alephium/explorer-backend/blob/42a0ab9e49c6bf2db3e271295ea4a73d780a0c30/app/src/main/scala/org/alephium/explorer/persistence/dao/BlockDao.scala#L110-L139) is using heavily the slick dsl.

This issue could be tackle at the same time as https://github.com/alephium/explorer-backend/issues/350 or not

PR #400, opened 2022-11-30
Partial #155: Raw SQL `updateMainChainStatus` and transactionally update multiple `blocks`

simerplaha:
- Partially towards #155
- Added test-cases. Existing test-cases were missing coverage for this `updateMainChainStatusAction` query.
- The query now accepts multiple blocks as input so `updateMainChain` could run on multiple blocks transactionally. I think this is needed for one of the other issues targeting sync optimisations.


tdroxler:
I did run some benchmarks and it all looks good, indeed it's nice to be able to update multiple blocks at once in the future :+1:

PR #404, opened 2022-12-08
ExecutionContext.reportFailure and remove all `sys.exit`

### LINKED ISSUE #395: Stop the app in case of an exception in PSQL or in our sync services
Currently if a fatal error happen in our sync services, the app won't stop. Those stateless services are scheduled by the `Scheduler` and on an error we only log, but then it get stuck.

We better stop the all app and let it restart.

tdroxler:
Resolves #395 and #246 

I took the opportunity to tackle the second as it's related.

`ExecutionContext.reportFailure` is now used to warn the top of the application of a failure, currently we stop the app on all event (we use `reportFailure` only once), but later we could choose to have different kind of behavior.

Now that `SyncServices` can warn and stop the app, I could remove the `sys.exit` from `BlockFlowSyncService`, just need to fail the future and it will be catch by `SyncServices` and trigger the `reportFailure`.

The latest `sys.exit` was in migration, but it was actually a dead branch that couldn't be reach.

following the code, `AppStateQuery.get(MigrationVersion)` couldn't return anything else than `None` or `Some(MigrationVersion(x))`, but from the compiler side, it thought it could return any `AppState`. So I rework a bit the `AppState` thing so the compiler understand and the `sys.exit` was automatically removed.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/Main.scala
+++ b/app/src/main/scala/org/alephium/explorer/Main.scala
@@ -29,6 +29,7 @@ import slick.basic.DatabaseConfig
 import slick.jdbc.PostgresProfile
 
 import org.alephium.explorer.config._
+import org.alephium.explorer.util.ExecutionContextUtil
 
 object Main extends StrictLogging {
   def main(args: Array[String]): Unit = {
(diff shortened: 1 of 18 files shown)


polarker (approved):
LGTM

PR #406, opened 2022-12-09
Feature/add input tx hash ref

tdroxler:
I'm still running the migration, but we can already start to review it.

In the second commit I change the migration to update inputs day by day, otherwise we can't really see the progress and we can think it stucks. 

EDIT: Migration done in around 1:30h, I checked multiple Inputs, data seems all correct. I also checked and all inputs where updated, (no inputs with `output_tx_hash_ref = nulll` in DB

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
@@ -29,6 +29,29 @@ import org.alephium.explorer.persistence.queries.AppStateQueries
 
 object Migrations extends StrictLogging {
 
+  def addInputTxHashRefColumn(): DBActionWT[Int] = {
+    sqlu"""
+      ALTER TABLE inputs
+      ADD COLUMN IF NOT EXISTS output_ref_tx_hash bytea;
+    """
+  }
+
+  //Will automatically be recreated after the migration
+  def dropInputOutputRefAddressNullIndex(): DBActionWT[Int] = {
+    sqlu"DROP INDEX IF EXISTS inputs_output_ref_amount_null_idx"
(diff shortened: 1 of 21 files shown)


app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
```diff
@@ -29,6 +29,29 @@ import org.alephium.explorer.persistence.queries.AppStateQueries
...
 object Migrations extends StrictLogging {
 
+  def addInputTxHashRefColumn(): DBActionWT[Int] = {
+    sqlu"""
+      ALTER TABLE inputs
+      ADD COLUMN IF NOT EXISTS output_ref_tx_hash bytea;
+    """
+  }
+
+  //Will automatically be recreated after the migration
```
polarker, line 44: Why don't use `create IF NOT EXISTS` for the index?

polarker (commented):
LGTM, merging for Gate.

polarker (approved):

app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
tdroxler: Index is created anyway here https://github.com/alephium/explorer-backend/blob/805da670f630e44f2e6550ba6bbf44aba241227f/app/src/main/scala/org/alephium/explorer/persistence/DBInitializer.scala#L100

Which lead to here https://github.com/alephium/explorer-backend/blob/805da670f630e44f2e6550ba6bbf44aba241227f/app/src/main/scala/org/alephium/explorer/persistence/schema/InputSchema.scala#L78-L81

Issue #398, opened 2022-11-24
`Fetch block` is still using the node addresses from the self-clique info 


tdroxler:
When tackling [this issue](https://github.com/alephium/explorer-backend/issues/214) 

as reported by @Lbqds : 

The code, follow this call chain:
BlockFlowSyncService.getRemoteMaxTimestamp
BlockFlowClient. fetchBlocksAtHeight
BlockFlowClient.fetchBlock
and the BlockFlowClient.fetchBlock fetches the full node address via infos/self-clique

I guess we need to pass the `direct-clique-access` to the `BlockFlowClient`

PR #409, opened 2022-12-12
Respect `direct-clique-access` flag in `BlockFlowClient`

tdroxler:
resolves: #398 

I checked other part of the codes to see if we need to also apply this patter, but it seems not.


Lbqds (approved):
LGTM 🎉

polarker (approved):

Issue #391, opened 2022-11-18
`/utils/update-global-loglevel` not working properly


tdroxler:
When starting with log level set to `DEBUG`, the endpoint works well, you can set `INFO` and go back to `DEBUG` correctly, the logs adapt well.

If you start with log level to `INFO`, calling the endpoint with `DEBUG` doesn't work, debug logs are never printed.

PR #410, opened 2022-12-13
Fix console logging level

tdroxler:
We were not configuring the log level at the right place, the threshold is a fixed limit that cannot be passed, it makes sense for the various `FileAppender`, but not for console (in our case).

Resolves: https://github.com/alephium/explorer-backend/issues/391

Calling the update-log-level endpoint now works as expected, I manually tested it.

I tried to add a Unit test, but using mocking is a nightmare and nothing is working or it's ugly as hell.

 @polarker let me know if you want me  to do the same in the full node.

### DIFF
--- a/app/src/main/resources/logback.xml
+++ b/app/src/main/resources/logback.xml
@@ -69,10 +69,6 @@
     </if>
 
     <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
-        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-            <level>${EXPLORER_LOG_LEVEL:-INFO}</level>
-        </filter>
-
         <encoder>
             <pattern>${consolePattern}</pattern>
         </encoder>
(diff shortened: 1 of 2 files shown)


app/src/main/resources/logback.xml
```diff
@@ -84,7 +80,7 @@
     <logger name="io.netty" level="WARN"/>
     <logger name="org.asynchttpclient" level="WARN"/>
 
-    <root>
+    <root level="${EXPLORER_LOG_LEVEL:-INFO}">
```
polarker, line 83: Does this change the log level of `ERRFILE` too?

tdroxler: no because the `ERRFILE` has a `Threshold` : https://github.com/alephium/explorer-backend/blob/cda9dae28be07a06c7274bd221d1edb840ddaa9f/app/src/main/resources/logback.xml#L16-L18

polarker:
In general
1. I am not sure if it's a good idea to change log configuration by endpoints. By doing this, one usually breaks the workflow of log infrastructure.
2. The log design of full node is that console logs and error logs should be fixed to use `Info` and `Error`. Only the `info` is configurable. The reason is that `console` should not be used for serious stuff and should be turn off in production.

tdroxler:
configurable log level through endpoint was discussed here https://github.com/alephium/explorer-backend/issues/196

polarker:
Thanks for the link. I think the configuration is fine for the full node, as we don't feel the need to change log level through endpoints for now.

This PR itself LGTM 👍

PR #411, opened 2022-12-16
Fix missing finalized outputs

tdroxler:
[TimeUtil.buildTimeStampRange](https://github.com/alephium/explorer-backend/blob/de891687eaa0fb11b1002bd61b49e77b41c90a90/app/src/main/scala/org/alephium/explorer/util/TimeUtil.scala#L52-L73): From this function, [it implies that our queries have to be inclusive](https://github.com/alephium/explorer-backend/blob/de891687eaa0fb11b1002bd61b49e77b41c90a90/app/src/test/scala/org/alephium/explorer/util/TimeUtilSpec.scala#L81-L82) on both `fromTs` and `toTs`: if our time ranges are: `(0,2), (3,5)`, if our query has `< toTs`, the first range will take only `0 -> 1` timestamps. On the second range it will take `3 -> 4`, So `2` and `5` will be ignored. We were also handling the case where the last time range was supposed to be exactly the same value, we were adding `+1ms` (see [test](https://github.com/alephium/explorer-backend/compare/fix/time-ranges?expand=1#diff-b44f35509058325905b061f74f32809f679f3f4b9698e3b2d40a6a50a14c0d90L79)), which is then wrong then when doing the queries, as we want both inclusive values, `<= toTs` was then taking value `+1ms` too much compared to what we asked for.

We will now see that every functions using this `buildTimeStampRange` is fine using inclusion from both side.

`FinalizerService`: [we need to have <=](https://github.com/alephium/explorer-backend/compare/fix/time-ranges?expand=1#diff-67c243d4b5cf51a0ac4a1913ea8f3ed48a022f151bc7d3a73c4e4e7a3e6ddc36R111) other wise on each round, the `to` will be missed. Also [previously we were not returning any time range if there were only 1 input](https://github.com/alephium/explorer-backend/compare/fix/time-ranges?expand=1#diff-67c243d4b5cf51a0ac4a1913ea8f3ed48a022f151bc7d3a73c4e4e7a3e6ddc36L132), it's really a corner case, but we still need it. If someone run a Devnet and do only one transaction, the output ref needs to marked as `spent`.

[BlockFlowSyncService](https://github.com/alephium/explorer-backend/blob/53f25a344d29555e1810bb53a9de33ceb8dcbe00/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L188):  all good to have inclusive `from - to` as [it's the case when fetching the node](https://github.com/alephium/alephium/blob/9b103b488d786c8933f41d9b75d43b10e42edb42/flow/src/main/scala/org/alephium/flow/core/BlockChain.scala#L138)

Migrations: [InputUpdateQueries.updateInputs](https://github.com/alephium/explorer-backend/blob/de891687eaa0fb11b1002bd61b49e77b41c90a90/app/src/main/scala/org/alephium/explorer/persistence/queries/InputUpdateQueries.scala#L35-L51) query doesn't care about timestamps, we used timestamp in the migrations to be able to log progress. So those who already ran the migration and might have miss some inputs, they will be eventually updated by the `updateInputs` function.

I ran the migrations with our latest archive, there were around 2300 outupts to update, took around 45s.

For all the test and checkup I did, I pretty sure only this `spent_finalized` was impacted

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/SyncServices.scala
+++ b/app/src/main/scala/org/alephium/explorer/SyncServices.scala
@@ -50,6 +50,7 @@ object SyncServices extends StrictLogging {
         Future.unit
 
       case BootMode.ReadWrite | BootMode.WriteOnly =>
+        logger.info("Starting sync services")
         getPeers(
           networkId          = config.networkId,
           directCliqueAccess = config.directCliqueAccess,
--- a/app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
@@ -28,11 +28,15 @@ import org.alephium.explorer.foldFutures
 import org.alephium.explorer.persistence.model.AppState.MigrationVersion
 import org.alephium.explorer.persistence.queries.AppStateQueries
 import org.alephium.explorer.persistence.schema.CustomSetParameter._
+import org.alephium.explorer.service.FinalizerService
 import org.alephium.explorer.util.TimeUtil
 import org.alephium.protocol.ALPH
 import org.alephium.util.{Duration, TimeStamp}
 
+@SuppressWarnings(Array("org.wartremover.warts.AnyVal"))
 object Migrations extends StrictLogging {
+
+  val latestVersion: MigrationVersion = MigrationVersion(3)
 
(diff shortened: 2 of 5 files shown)


polarker (approved):
This is indeed a tricky case. Well done ❤️

Issue #253, opened 2022-05-30
Add examples to openapi.yaml


tdroxler:
let's give nice examples to make our openapi file a bit nicer and easier to understand.

PR #412, opened 2023-01-05
#253: OpenAPI `Endpoint` examples for Block APIs

simerplaha:
- Towards #253.
- Added `Example`s for `BlockEndpoints`.
- Everything is declared as `val` in `EndpointExamples` following the node. Assuming that's preferred.


tdroxler (approved):
:rocket: good point for the `val` I'll have a look, in the mean time let's go with them, it's easy to change later.

PR #414, opened 2023-01-06
#253: Some more OpenAPI examples

simerplaha:
- Towards #253
- Updated endpoint `listUnconfirmedTransactions` to return `UnconfirmedTransaction` instead of `TransactionLike`

Kinda need some help making sure that the default values in the `Example`s make sense.  

Examples are added for `jsonBody` only. Should we do this for `plainBody` as well?


app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
```diff
@@ -172,6 +172,38 @@ object EndpointExamples extends EndpointsExamples {
...
+  private val explorerInfo =
+    ExplorerInfo(
+      releaseVersion = "1.11.2+17-00593e8e-SNAPSHOT",
+      commit         = "00593e8e8c718d6bd27fe218e7aa438ef56611cc"
+    )
+
+  private val tokenSupply =
+    TokenSupply(
+      timestamp   = ts,
+      total       = U256.MaxValue,
```
tdroxler, line 185: using MaxValue is a bit too much, let's maybe divide by 10^x to reach a level of magnitude like in https://mainnet-backend.alephium.org/infos/supply

app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
```diff
@@ -186,7 +186,7 @@ object EndpointExamples extends EndpointsExamples {
       circulating = U256.HalfMaxValue,
       reserved    = U256.Ten,
       locked      = U256.Ten,
-      maximum     = U256.Ten
+      maximum     = U256.MaxValue
```
tdroxler, line 189: [let's use](https://github.com/alephium/explorer-backend/blob/f1697856dc0a600d04b93583b1ddb4fdb3fb3e0f/app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala#L124)

app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
```diff
@@ -204,6 +205,29 @@ object EndpointExamples extends EndpointsExamples {
...
+        "  Filter: table_column",
+        "Planning Time: 0.694 ms",
+        "Execution Time: 5.432 ms"
+      ),
+      messages = Array(
+        "Used table_column_idx = false",
+        "Used table_pk         = true"
+      ),
+      passed = true
+    )
```
tdroxler, line 223: it's weird it doesn't show int the  openapi.json.

tdroxler: oh it's because it's not in `Documentation`, but still it's nice to have the example in the code.

app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
simerplaha: I don't how to divide an Integer with a variable function. 10^x is a function. What is value of `x` here?

Please explain.

tdroxler: something like: 
```
ALPH.MaxALPHValue.divUnsafe(U256.Billion) = 1000000000000000000
```
so x is 9 in that case: 1 billion = 10^9

the idea is just to make the example more friendly to read, with `U256.MaxValue` it's a bit scary: `115792089237316195423570985008687907853269984665640564039457584007913129639935` :sweat_smile:

app/src/main/scala/org/alephium/explorer/api/UnconfirmedTransactionEndpoints.scala
```diff
@@ -32,9 +32,9 @@ trait UnconfirmedTransactionEndpoints extends BaseEndpoint with QueryParams {
       .tag("Transactions")
       .in("unconfirmed-transactions")
 
-  val listUnconfirmedTransactions: BaseEndpoint[Pagination, ArraySeq[TransactionLike]] =
+  val listUnconfirmedTransactions: BaseEndpoint[Pagination, ArraySeq[UnconfirmedTransaction]] =
```
tdroxler, line 35: I know it makes more sense to have `UnconfirmedTransaction` here, but then tapir is generating a new schema:

```UnconfirmedTransaction1```

which is not nice when front end is auto-generating their type-script files, having this `1` is pretty ugly.

Until I come with a better idea, let's keep `TransactionLike` to avoid creating a new schemas

File: app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala (hunk 1 above)
Comment (simerplaha) [reply app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala:185]: Should `circulating` should be changed as well? `U256.HalfMaxValue` also looks a bit scary 😅 

```scala
circulating = ALPH.MaxALPHValue.divUnsafe(U256.Billion).divUnsafe(U256.Two)
```

```json
"total"      : "1000000000000000000",
"circulating": "500000000000000000"
```

simerplaha:
> I haven't thought about it, but yes, so we can be explicit in the example that the plain body is actually a number or a string.

Nice. Can look at `plainBody` stuff in another PR?

tdroxler:
> Nice. Can look at `plainBody` stuff in another PR?

sure, let's have a single PR for that

PR #415, opened 2023-01-09
Feature/coinbase flag

### LINKED ISSUE #379: Add a `coinbase` flag in transaction table
It could ease if we want to do queries on non-coinbase txs, for example to create a chart of all non-coinbase txs.

tdroxler:
Resolves #379

I had this idea few weeks ago and did some experiment in this branch. As [this PR](https://github.com/alephium/explorer-backend/pull/376) will need a resync, it could be a nice time to introduce those new fields.

Not sure it's 100% needed fields, but I had this idea while preparing some stats for our 1 year anniversary, we couldn't easily query how many non-coinbase txs we have, as it require an heavy join on the inputs. With such flag, it would become quite easy.


simerplaha (commented):
Nice 👍🏼

Just wondering if it makes it easy to add more exchanges in the future.

Following the current approach of using a `coinbase: Boolean` column flag to represent a coinbase transaction and `coinbase_tx_id: TransactionId` to represent a coinbase transaction id - if we were to add more exchanges in the future we would need more `boolean` and `transaction_id` columns for each exchange which might add schema migration work and the table might get large with a lot of redundant flags with majority of columns set to `false`.

What do you think about inverting it?

Instead of the table column `coinbase: Boolean`, we could use 
- `exchange_type: String` and have `exchange_type = 'coinbase'` 
-  or `exchange_type: Byte` where `0` could represent `coinbase`?

tdroxler:
> if we were to add more exchanges in the future 

`coinbase` here is not referring to the [exchage](https://www.coinbase.com), but to a [coinbase transaction](https://academy.bit2me.com/en/que-es-coinbase-transaccion/) which is a special type of transaction that are creating new coins, so such transactions don't have inputs.

This is interesting to have such flag for example to query how many txs are actually done "for real": people sending alph to another address.

But still it's a good point, could we have in the future other category of transactions @polarker ? Rather than just coinbase vs non-coinbase txs?

simerplaha:
> coinbase here is not referring to the [exchage](https://www.coinbase.com/), but to a [coinbase transaction](https://academy.bit2me.com/en/que-es-coinbase-transaccion/) which is a special type of transaction that are creating new coins, so such transactions don't have inputs.

Oh whoops! Learnt something new. Cheers to that :)

polarker (approved):
LGTM, but I can only quickly review it

Issue #413, opened 2023-01-05
Investigate better index for our timestamps


tdroxler:
Based on [this](https://www.postgresql.org/docs/current/indexes-ordering.html)
We should maybe consider adding this `DESC` in some of our timestamp indexes, as most of the time we order them using `DESC`

tdroxler:
hey @simerplaha need your expertise here. What's your opinion on that?

simerplaha:
Oh, good catch! 👍🏼 

You are right. Sounds like we should add explicit `DESC` ordering on some of our timestamp indexes.

From the doc I'm understanding that for single column queries we might not need explicit `DESC` ordering because the default `ASC` satisfies scans in both directions. If I'm misreading something, a query planner check should clear this. But I guess it wouldn't hurt to make them all explicitly `DESC`.

We do need to add indexes and explicit ordering for the following multi-column `order by` queries:

| Table                       | `ORDER BY`                                | Queries                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
|-----------------------------|-------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `transaction_per_token`     | `ORDER BY block_timestamp DESC, tx_order` | [query](https://github.com/alephium/explorer-backend/blob/6610288e6b05a57ffc80dc03cb7277e26f1938b2/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L103)                                                                                                                                                                                                                                                                                                                                                                                                        |
[...]

Issue #205, opened 2022-04-22
Improve multi-column index


polarker:
When the multi-column index is of b-tree type, we'd better put the most important index in the first place.

tdroxler:
To help this, we can run first run the following query, that shows btree indexes that cover more than one column.

```sql
SELECT
    t.relname AS table_name,
    i.relname AS index_name,
    array_to_string(array_agg(a.attname ORDER BY x.n), ', ') AS columns
FROM
    pg_class t
    JOIN pg_index ix ON t.oid = ix.indrelid
    JOIN pg_class i ON i.oid = ix.indexrelid
    JOIN pg_am am ON i.relam = am.oid
    JOIN LATERAL unnest(ix.indkey) WITH ORDINALITY AS x(attnum, n) ON true
    JOIN pg_attribute a ON t.oid = a.attrelid AND a.attnum = x.attnum
WHERE
    t.relkind = 'r'
    AND i.relkind = 'i'
    AND am.amname = 'btree'
GROUP BY
    t.relname, i.relname
HAVING
    COUNT(*) > 1
ORDER BY
    t.relname, i.relname;
```

Checking then every queries that use those indexes will be another heavy task.

PR #416, opened 2023-01-13
#413: Index updates to `transaction_per_token`

simerplaha:
Towards #413 & #205

Added multi-column index for the table `transaction_per_token` used by [this query](https://github.com/alephium/explorer-backend/blob/fda63896eb59f16bd33bfc64db099712a7f391b2/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L103).

```sql
(block_timestamp desc, tx_order asc); -- multi-column index
(block_timestamp desc); -- updated existing index from asc to desc
```

We can remove the single-column index `(block_timestamp desc)` because the multi-column index `(block_timestamp desc, tx_order asc)` serves the same purpose. But I’m not able to confirm this because the query planner sometimes (at least initially) prefers to use `(block_timestamp desc)`. We also don’t have any queries that runs just on `block_timestamp`. I’d personally remove the single-column index but there is no harm in leaving it in. 

## Overview 

Here is a very high level overview of what I understand after running multiple tests using [this query](https://github.com/alephium/explorer-backend/blob/fda63896eb59f16bd33bfc64db099712a7f391b2/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L103) to see how the `multi-column index` behaves on different sized data (1,000 - 500,000 tokens) which might also apply to other tables mentioned in the [issue](https://github.com/alephium/explorer-backend/issues/413#issuecomment-1376714679) when adding these `multi-column indexes`.

The `multi-column index` gets used
- when the `token_id` being queried has large number of rows (tested with 10 tokens, 50,000 rows per token)
- and at an `OFFSET` away from the head or tail-end. 

But if the `token_id` has small number of rows (say `10` rows) it just uses `token_idx`.

A query ordering by `block_timestamp desc` does a backward scan on an index with ascending order (`block_timestamp_index asc`) . If the index is in descending order (`block_timestamp_index desc`) it does forward scan. The query planner showed no noticeable difference in execution time for either one of them. <strike>I didn’t do any JMH benchmarks.</strike> Benchmarks added below.

From my own experience, storage engines have a small performance cost running backward scan over forward scan. So I'd also prefer having `(block_timestamp desc)` to avoid backward scans. In storage engines there is also a small storage cost (a reference) for backward scan. But I didn’t see any extra cost in Postgres, both indexes weigh the same `11255808 bytes (11.25 mb)`.


simerplaha:
Ran some quick JMH benchmarks in a different [branch](https://github.com/alephium/explorer-backend/tree/413_multi_column_index_transaction_per_token_bench). Referenced in the last commit here. 

The multi-column index does better. The query planner output is also convincing. 

Pasted Graphic 1

tdroxler:
Thx as always for the benchmarks, the differences aren't huge, but still it's better with the multi-index + it seems to make more sense to have such index based on our queries. 
What's your opinion on that @polarker ?

polarker:
LGTM too

polarker (approved):

tdroxler (approved):

PR #419, opened 2023-01-17
Convert `UnlockScript.SameAsPrevious` to the correct value + Bump common to 1.6.x

tdroxler:
The  idea is to convert directly the protocol inputs as soon as we access them (only twice in the code) to be sure we have the correct unlock scripts latter.


polarker (approved):
LGTM

PR #421, opened 2023-01-18
#413: multi-column timestamp indexes

simerplaha:
Towards #413.

Is there a need to maintain the index names as before?

- Table named `token_tx_per_addresses` has indexes prefixed as `token_tx_per_address_*`
- Table named `transaction_per_addresses` has indexes prefixed as `txs_per_address_*`

This PR use the same prefix as table name.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/CommonIndex.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/CommonIndex.scala
@@ -30,9 +30,9 @@ object CommonIndex {
     * Postgres uses this index for queries that order by `block_timestamp desc, tx_order asc`
     * and have large number of resulting rows.
     */
-  def blockTimestampTxnOrderIndex(table: Schema[_]): SqlAction[Int, NoStream, Effect] =
+  def blockTimestampTxOrderIndex(table: Schema[_]): SqlAction[Int, NoStream, Effect] =
     sqlu"""
-        create index if not exists #${table.name}_block_timestamp_txn_order_idx
+        create index if not exists #${table.name}_block_timestamp_tx_order_idx
                 on #${table.name} (block_timestamp desc, tx_order asc);
         """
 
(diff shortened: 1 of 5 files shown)


app/src/main/scala/org/alephium/explorer/persistence/schema/CommonIndex.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+    * Need for this multi-column index?
+    *
+    * Postgres uses this index for queries that order by `block_timestamp desc, tx_order asc`
+    * and have large number of resulting rows.
+    */
+  def blockTimestampTxnOrderIndex(table: Schema[_]): SqlAction[Int, NoStream, Effect] =
+    sqlu"""
+        create index if not exists #${table.name}_block_timestamp_txn_order_idx
```
tdroxler, line 35: `s/_block_timestamp_txn_order_idx/_block_timestamp_tx_order_idx` ?

so `s/txn/tx`

app/src/main/scala/org/alephium/explorer/persistence/schema/TransactionPerTokenSchema.scala
```diff
@@ -47,26 +46,11 @@ object TransactionPerTokenSchema
...
-        create index if not exists #${name}_timestamp_idx
-                on #$name (block_timestamp desc);
-        """
-
   def createSQLIndexes(): DBIO[Unit] =
-    DBIO.seq(timestampTxnOrderIndex(), timestampIndex())
+    DBIO.seq(
+      CommonIndex.blockTimestampTxnOrderIndex(this),
+      CommonIndex.timestampIndex(this)
+    )
```
tdroxler, line 53: great idea the `CommonIndex`

tdroxler:
>Is there a need to maintain the index names as before?

nope, I was probably lazy ^^ but better to do as you propose to use `table.name` directly.

tdroxler (approved):

app/src/main/scala/org/alephium/explorer/persistence/schema/CommonIndex.scala
simerplaha: Oh yep. Didn't realise `tx` is consistent with the rest of the code.

Just to confirm is this suggesting that I use `tx` instead of `txn`?

tdroxler: yep

PR #424, opened 2023-01-19
Coverage: Test cases for `utransactions` queries 

simerplaha:
Implements test cases for queries running on table `utransactions`.

## Coverage 

| Package: `org.alephium.explorer.persistence.queries` | Class | Method | Line |
|------------------------------------------------------|----------------|-----------------|---------------|
| Existing  coverage                                  | 72% (29/40)    | 59% (203/343)   | 67% (640/950) | 
| Updated coverage                                    | 77% (31/40)    | 61% (212/343)   | 70% (665/950) |


app/src/test/scala/org/alephium/explorer/api/model/PaginationSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package org.alephium.explorer.api.model
+
+import org.alephium.explorer.AlephiumFutureSpec
+import org.alephium.explorer.Generators._
+import org.alephium.explorer.persistence.{DatabaseFixtureForEach, DBRunner}
+
+class PaginationSpec extends AlephiumFutureSpec with DatabaseFixtureForEach with DBRunner {
+
+  "pagination" should {
```
tdroxler, line 25: I would maybe precise here: `paginationGen` because we are testing the generator here, not the actual `Pagination`

app/src/test/scala/org/alephium/explorer/persistence/queries/UnconfirmedTransactionQueriesSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.alephium.explorer.AlephiumFutureSpec
+import org.alephium.explorer.Generators._
+import org.alephium.explorer.persistence.{DatabaseFixtureForEach, DBRunner}
+import org.alephium.explorer.persistence.model._
+import org.alephium.explorer.persistence.schema._
+import org.alephium.util.TimeStamp
+
+class UnconfirmedTransactionQueriesSpec
+    extends AlephiumFutureSpec
+    with DatabaseFixtureForEach
```
tdroxler, line 33: I think you could use `DatabaseFixtureForAll` here, since you are anyway deleting the data every time

app/src/test/scala/org/alephium/explorer/api/model/PaginationSpec.scala
simerplaha: Oh you are right. I will just remove the test yea? Doesn't make sense to be testing a generator :)

tdroxler (approved):
Minor comments, but LGTM :rocket:

app/src/test/scala/org/alephium/explorer/api/model/PaginationSpec.scala
tdroxler: yeah let's remove it

PR #429, opened 2023-02-10
Use `protocol.model.Address` instead of our custom one

tdroxler:
In the beginning of the `explorer-backend`, addresses were just used to be displayed.

With the current implementation of the tokens and the future one with contracts, we start to use more heavily the addresses. Having access to all the function coming from the protocol type is quite convenient.

We'll also be less error prone if the `protocol.model.Address` type is changing in the future.

We could probably do the same for `Height/GroupIndex` etc, trying to remove duplicates between the two code base.


simerplaha (approved):
Nice 🚀

Issue #427, opened 2023-02-08
`GET:/addresses/{address}/unconfirmed-transactions` returns `TransactionLike[]` and not  UnconfirmedTransaction[]?


tdroxler:
`GET:/addresses/{address}/unconfirmed-transactions` returns `TransactionLike[]` and not `UnconfirmedTransaction[]`

`UnconfirmedTransaction` is extending `TransactionLike` and thus has a `type` field in his JSON schema.
If we try to return directly `UnconfirmedTransaction` in an endpoint, `tapir` doesn't understand we want the same schema and is creating a new `UnconfirmedTransaction1` schema in the `openapi.yaml`, withouth the `type` field.

It would be nicer if the above endpoint in the `openapi.yaml` could return the `UnconfirmedTransaction` with the `type` field. So we could have only 1 schema for it.

The other solution is to define a new type, the same way we do the separation of `Transaction` and `ConfirmedTransaction`. But we need to think of a good naming to avoid confusion.

nop33:
`MempoolTransaction`? 😅

tdroxler:
could be an idea yes, but then we should maybe rename the endpoint too?

nop33:
Just to verify, unconfirmed transactions are the ones that have not yet been included in a block? They are just in the mempool, awaiting to be added in a block. Right?

tdroxler:
yes exactly

nop33:
Since that's the case, `MempoolTransaction` and `GET:/addresses/{address}/mempool-transactions` sounds good to me 👍 

I'll let @mvaivre give his opinion as well.

tdroxler:
but then it means that `TransactionLike` will still be `ConfirmedTransaction` and `UnconfirmedTransaction`

nop33:
imo, this could be a better naming convention:
```ts
type Transaction = ConfirmedTransaction | UnconfirmedTransaction
```

tdroxler:
ahah but `Transaction` is already taken to avoid having `Transaction1` :stuck_out_tongue_closed_eyes:  anyway I need to figure how to go around that issue in `tapir`

mvaivre:
Yeah, basically the solution is to solve the issue in `tapir` otherwise we'll go around in circles 😝 (ie. using a new `MempoolTransaction` type will make things even more confusing if `TransactionLike` is still `ConfirmedTransaction | UnconfirmedTransaction`).

[...]

PR #431, opened 2023-02-20
Rename `unconfirmed-transaction` to `mempool-transaction`/

tdroxler:


polarker:
I am not sure about `ConfirmedTransaction`, because it's not confirmation if a transaction is included into a new block. It usually takes hours to confirm at the current stage.

How about using `Transaction` + `MempoolTransaction` instead of `ConfirmedTransaction`? Since the trait is `TransactionLike`, this naming should work?

simerplaha (approved):
🚀

tdroxler:
> I am not sure about `ConfirmedTransaction`, because it's not confirmation if a transaction is included into a new block. It usually takes hours to confirm at the current stage.
> 
> How about using `Transaction` + `MempoolTransaction` instead of `ConfirmedTransaction`? Since the trait is `TransactionLike`, this naming should work?

We already have `Transaction`, we duplicated `ConfirmedTransaction` otherwise tapir was creating a `Transaction1` schema depending if it was used as a `Transaction` or in the `TransactionLike` context. This was also raised in [that issue](https://github.com/alephium/explorer-backend/issues/427). I'll spend a bit of time on this, I'm pretty sure we can tweak the openapi generation.

tdroxler:
@polarker I pushed the changes we discussed:

* Transaction
* MempoolTransaction
* TransactionLike = AccpetedTransaction | PendingTransaction

polarker (approved):
LGTM. Does it well with Typescript now?

Issue #426, opened 2023-01-25
Update queries spliced literal values to be parametric queries


simerplaha:
The following queries are using spliced literal values (`#$param`). They can be updated to be parameterised queries so they are used as Prepared Statements better.

# Paginated queries

This is a relatively recent change so want to confirm if this is something we'd want changed?

```scala
#${pagination.query}
```

The problem here is that because [Pagination.limit](https://github.com/alephium/explorer-backend/blob/56f8d97990a6eb99a86b8a8b37951b13d59b975b/app/src/main/scala/org/alephium/explorer/api/model/Pagination.scala#L21) is being embedded in the following paginated queries as `String` value instead of query parameters, this reduces the chance of these queries being used as Prepared Statements for cases where `limit` and `offset` differ between subsequent queries. So locally the client will not be able to cache these queries well & will have to recompile and also the server will also not be able to cache 'em well.

- [TokenQueries.listTokensAction](https://github.com/alephium/explorer-backend/blob/56f8d97990a6eb99a86b8a8b37951b13d59b975b/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L71)
- [TokenQueries.getAddressesByToken](https://github.com/alephium/explorer-backend/blob/56f8d97990a6eb99a86b8a8b37951b13d59b975b/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L88)
- [TokenQueries.listTokenTransactionsAction](https://github.com/alephium/explorer-backend/blob/56f8d97990a6eb99a86b8a8b37951b13d59b975b/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L100)
- [TokenQueries.getTokenTxHashesByAddressQuery](https://github.com/alephium/explorer-backend/blob/56f8d97990a6eb99a86b8a8b37951b13d59b975b/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L130)
- [TransactionQueries.getTxHashesByBlockHashWithPaginationQuery](https://github.com/alephium/explorer-backend/blob/56f8d97990a6eb99a86b8a8b37951b13d59b975b/app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala#L141)
[...]

PR #434, opened 2023-02-24
Sub-contract creation event

tdroxler:
The goal of this PR is to tackle [this comment](https://github.com/alephium/token-list/pull/4#issuecomment-1418693207). 

Not sure yet about the different naming, but it comes from the full node and the hardcoded [createContractEventIndex](https://github.com/alephium/alephium/blob/master/protocol/src/main/scala/org/alephium/protocol/vm/package.scala#L47)

[A contract write a log on creation](https://github.com/alephium/alephium/blob/c84dc660b8e7648827d16c3caef78f87205405fc/protocol/src/main/scala/org/alephium/protocol/vm/Frame.scala#L373-L408) from the `contractCreationEventFields` function, we know that the first address in the log fields is the sub-contract address, while the second address is the parent one.
We can ignore the first field `createContractEventIndex`, as it's discarded [here](https://github.com/alephium/alephium/blob/c84dc660b8e7648827d16c3caef78f87205405fc/protocol/src/main/scala/org/alephium/protocol/vm/event/MutableLog.scala#L40).

In the explorer-backend, when we receive and event, we can check if it's a contract creation event with the `createContractEventIndex` contract address and then check if there are exactly 2 fields, which means it's a sub-contract creation as defined by then node, we need to be careful in the future if we change the log in the node, for example if we add a field or reverse the two addresses.

I still need to write a migration for this, so we don't need to re-sync from scratch. As all events are already stored in db, it should be to hard, as we can loop on all our events where contractId = createContractEventIndex.


simerplaha (commented):
Nice.

h0ngcha0:
Perhaps it makes sense to be able to get a list of subcontract ids given its parent id, the use case would be displaying all NFTs for an NFT collection. WDTY?

ping @polarker as well

app/src/test/scala/org/alephium/explorer/persistence/queries/ContractQueriesSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+      val parent     = addressGen.sample.get
+      val pagination = Pagination.unsafe(1, 5)
+
+      forAll(Gen.nonEmptyListOf(createEventGen(Some(parent))), Gen.nonEmptyListOf(createEventGen())) {
+        case (events, otherEvents) =>
+          run(CreateSubContractEventSchema.table.delete).futureValue
+          run(ContractQueries.insertSubContractCreation(events ++ otherEvents)).futureValue
+
+          run(ContractQueries.getSubContractsQuery(parent, pagination)).futureValue is events
+            .sortBy(_.timestamp)
```
h0ngcha0, line 94: [Nit] Is it possible to test the order inside of the block too?

app/src/main/scala/org/alephium/explorer/api/model/ContractParent.scala
```diff
@@ -0,0 +1,30 @@
...
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+package org.alephium.explorer.api.model
+
+import org.alephium.explorer.api.Json._
+import org.alephium.json.Json._
+import org.alephium.protocol.model.Address
+
+@SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+final case class ContractParent(
```
h0ngcha0, line 24: Question: would `Option[Address]` be enough? :P

h0ngcha0: Okey, I was thinking about a more Resty way: returning `Address` when the parent exists and 404 when the parent doesn't, and always returning a list of `Address` for subcontracts.
But it is more important to be consistent with other endpoints, if it is the similar approach that it is good :+1:

h0ngcha0 (commented):
LGTM, only minor comments and questions :+1:

app/src/main/scala/org/alephium/explorer/api/model/ContractParent.scala
tdroxler: 404 is when the contract itself doesn't exist, if you get `None` you know the contract exists but has no parent.

app/src/test/scala/org/alephium/explorer/persistence/queries/ContractQueriesSpec.scala
tdroxler: this test have 1 event per block, I'll try to add a test for that :+1:

app/src/test/scala/org/alephium/explorer/persistence/model/CreateSubContractEventEntitySpec.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.alephium.explorer.GenApiModel._
+import org.alephium.explorer.GenDBModel._
+import org.alephium.explorer.persistence.model.CreateSubContractEventEntity
+
+class CreateSubContractEventEntitySpec extends AlephiumSpec {
+
+  "CreateSubContractEventEntity.fromEventEntity" should {
+    "return None for a random event" in {
+      forAll(eventEntityGen) { event =>
+        CreateSubContractEventEntity.fromEventEntity(event) is None
```
polarker, line 32: `eventEntityGen` might generate a valid subcontract event with low probability or not?

app/src/main/scala/org/alephium/explorer/persistence/queries/ContractQueries.scala
```diff
@@ -0,0 +1,88 @@
...
+      """.asASE[Address](addressGetResult).headOrNone
+  }
+
+  def getSubContractsQuery(contract: Address, pagination: Pagination): DBActionSR[Address] = {
+    sql"""
+      SELECT sub_contract
+      FROM create_sub_contract_events
+      WHERE contract = $contract
+      ORDER BY block_timestamp DESC, event_order_in_block
+      #${pagination.query}
```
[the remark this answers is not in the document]
polarker: > Spliced pagination here would affect the query's ability to be used as prepared statement.

Great catch! Is it possible to reproduce this in a unit test?

app/src/main/scala/org/alephium/explorer/persistence/model/CreateSubContractEventEntity.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.alephium.util.TimeStamp
+
+final case class CreateSubContractEventEntity(
+    blockHash: BlockHash,
+    txHash: TransactionId,
+    contract: Address,
+    subContract: Address,
+    timestamp: TimeStamp,
+    eventOrder: Int
+)
```
polarker, line 31: Should we change this as:
```
  parentContract: Option[Address],
  newContract: Address,
```

In this way, we could store all of the contract creation events, not just for subcontracts.
Later, we could use the same database to query how many contracts created in total.

tdroxler: this is already stored in the [global event table](https://github.com/alephium/explorer-backend/blob/feature/contract-subcontract/app/src/main/scala/org/alephium/explorer/persistence/schema/EventSchema.scala). The idea here was to have a specific table for the sub-contract creation for fast queries.

tdroxler: oh but your idea was to have a table only for `Contract Creation` events?

polarker: Probably a table for `Contract` makes more sense. The fields could be:
```
Contract(
  address: Address, parentContract: Address,
  creationTS: TimeStamp, creationBlock: BlockHash, ..., // all creation columns
  destructionTs: TimeStamp, destructionBlock: BlockHash, ... // all destruction columns.
)
```

polarker: In the future, we may insert more data for each contract

app/src/main/scala/org/alephium/explorer/persistence/queries/ContractQueries.scala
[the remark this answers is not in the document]
tdroxler: good point, I miss your issue #426 , really nice one. I guess it could be tackle on it's own. 
The schema changed a bit with my latest commits, I'll add the various idx.

app/src/test/scala/org/alephium/explorer/persistence/model/CreateSubContractEventEntitySpec.scala
tdroxler: yes, but the `contractAddress` will never be `createContractEventId`

app/src/main/scala/org/alephium/explorer/persistence/queries/ContractQueries.scala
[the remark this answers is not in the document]
tdroxler: I did tried to add various indexes for timestamp, timestamp and event_order. But no one is ever used when performing the queries. 
As always with PG it depends with the amount of data, I tried on testnet but we still don't have too much contracts.
I think we could start without those indexes and retry in few monthes when we have more data

app/src/main/scala/org/alephium/explorer/persistence/queries/ContractQueries.scala
```diff
@@ -66,6 +81,36 @@ object ContractQueries {
     }
   }
 
+  private def updateContractDestructionEventEntities(
+      destroyInfos: Iterable[ContractEntity.DestroyInfo]): DBActionW[Int] = {
+    QuerySplitter.splitUpdates(rows = destroyInfos, columnsPerRow = 5) { (destroyInfos, _) =>
```
polarker, line 86: Does it skip DB insertion when the `Iterable` is empty?

polarker (approved):
LGTM, just one remark.

app/src/main/scala/org/alephium/explorer/persistence/queries/ContractQueries.scala
tdroxler: yes if it's empty DBAction is 
```
 DBIOAction.successful(0)
```
from [here](https://github.com/alephium/explorer-backend/blob/741f9996ee7d58f0bd859680603a0c1fce70b8e5/app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala#L41)
It's in [this part of the code](https://github.com/alephium/explorer-backend/blob/741f9996ee7d58f0bd859680603a0c1fce70b8e5/app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala#L122-L140) that if [the rows are empty ](https://github.com/alephium/explorer-backend/blob/741f9996ee7d58f0bd859680603a0c1fce70b8e5/app/src/main/scala/org/alephium/explorer/persistence/queries/QuerySplitter.scala#L123-L124) we return the above

polarker: Great.

PR #436, opened 2023-03-06
Add `Transaction.scriptExecutionOk` field

tdroxler:


app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
```diff
@@ -110,26 +110,28 @@ object EndpointExamples extends EndpointsExamples {
...
-      gasPrice  = org.alephium.protocol.model.nonCoinbaseMinGasPrice.value,
-      coinbase  = false
+      hash              = txId,
+      blockHash         = blockHash,
+      timestamp         = ts,
+      inputs            = ArraySeq(input),
+      outputs           = ArraySeq(outputAsset, outputContract),
+      gasAmount         = org.alephium.protocol.model.minimalGas.value,
+      gasPrice          = org.alephium.protocol.model.nonCoinbaseMinGasPrice.value,
+      scriptExecutionOk = false,
```
polarker, line 120: maybe use true as the default example.

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -351,24 +357,25 @@ object TransactionQueries extends StrictLogging {
...
+    val gasByTx = gases.groupBy(_.txHash).view.mapValues(_.map(_.info()))
     txHashesTs.map { txn =>
-      val ins                   = insByTx.getOrElse(txn.txHash, ArraySeq.empty)
-      val ous                   = ousByTx.getOrElse(txn.txHash, ArraySeq.empty)
-      val gas                   = gasByTx.getOrElse(txn.txHash, ArraySeq.empty)
-      val (gasAmount, gasPrice) = gas.headOption.getOrElse((0, U256.Zero))
+      val ins                                      = insByTx.getOrElse(txn.txHash, ArraySeq.empty)
+      val ous                                      = ousByTx.getOrElse(txn.txHash, ArraySeq.empty)
+      val gas                                      = gasByTx.getOrElse(txn.txHash, ArraySeq.empty)
+      val (gasAmount, gasPrice, scriptExecutionOk) = gas.headOption.getOrElse((0, U256.Zero, false))
```
polarker, line 365: I think by default it should be `true`, which means the transaction succedded at least. It will false only when the transaction failed.

polarker (approved):
LGTM

PR #450, opened 2023-04-04
Add `/addresses/<address>/tokens-balance` endpoint

tdroxler:
On the explorer front-end, we start to have `too many request` on address page with lots of tokens, as currently it needs to loop and call the balance endpoint for each token.

This endpoint allows to do only one call to get all tokens info needed.

This is currently quite a naive approach, also we need to handle the case where balance is 0, we shouldn't return the tokens in those case. But this requires some work to do things in an efficient way.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala
@@ -20,6 +20,7 @@ import scala.collection.immutable.ArraySeq
 import scala.concurrent.ExecutionContext
 
 import com.typesafe.scalalogging.StrictLogging
+import slick.dbio.DBIOAction
 import slick.jdbc.PostgresProfile.api._
 
 import org.alephium.explorer.api.model._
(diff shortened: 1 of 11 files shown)


app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala
```diff
@@ -110,6 +111,22 @@ object TokenQueries extends StrictLogging {
...
+  def listAddressTokensBalanceAction(address: Address, pagination: Pagination)(
+      implicit ec: ExecutionContext): DBActionSR[TokenBalance] =
+    for {
+      tokens <- listAddressTokensAction(address, pagination)
+      balances <- DBIOAction.sequence(tokens.map { token =>
+        getTokenBalanceAction(address, token).map { balance =>
+          (token, balance)
+        }
+      })
+    } yield {
```
simerplaha, line 123: Since the default limit here is `100`, this query could dispatch `100` queries for each request. Can we reduce this to fetch the data required in `1` or `2` queries?

If we don't wanna `join` then `getTokenBalanceAction` can query multiple tokens in a single query?

tdroxler: I confess that seeing our [getTokenBalance query](https://github.com/alephium/explorer-backend/blob/07c05f2e1e420b369359792db1dfd59293444b73/app/src/main/scala/org/alephium/explorer/persistence/queries/TokenQueries.scala#L47-L65) made me go for the simplest solution :sweat_smile: 
Nevertheless I did try to think of something going your way and I'm not sure how as we have to join on the inputs anyway. 
I'll give a new try.

PR #453, opened 2023-04-06
Make parametric query for pagination

tdroxler:
Moving toward https://github.com/alephium/explorer-backend/issues/426

@simerplaha I gave a try to replace the spliced query for pagination

I'm not 100% sure about it, so I need your expertise,  but I like how small the diff is (git extremist here) :)


simerplaha (approved):
Looks good. Since `SetParameter` and `action` has to be duplicated when `paginate` function is called, kinda would've preferred that pagination didn't get special treatment. But it looks good. It resolves the issue and very convenient :thumbsup:.

PR #454, opened 2023-04-14
Add parallelism to stream

tdroxler:
I initially thought that [mapAsync](https://doc.akka.io/api/akka/2.7/akka/stream/scaladsl/Source.html#mapAsync[T](parallelism:Int)(f:Out=%3Escala.concurrent.Future[T]):FlowOps.this.Repr[T]) was not preserving order, while it actually does. There is a specific [mapAsyncUnordered](https://doc.akka.io/api/akka/2.7/akka/stream/scaladsl/Source.html#mapAsyncUnordered[T](parallelism:Int)(f:Out=%3Escala.concurrent.Future[T]):FlowOps.this.Repr[T]) if you don't want to preserve.

This improved a lot the performance, 8 seems a good value while doing my test, going above didn't improved.
I added a config value so we can easily play with it if we want.

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -38,7 +38,7 @@
     </check>
     <check level="error" class="org.scalastyle.scalariform.ParameterNumberChecker" enabled="true">
         <parameters>
-            <parameter name="maxParameters"><![CDATA[8]]></parameter>
+            <parameter name="maxParameters"><![CDATA[9]]></parameter>
         </parameters>
     </check>
     <check level="error" class="org.scalastyle.scalariform.MagicNumberChecker" enabled="true">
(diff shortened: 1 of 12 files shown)


scalastyle-config.xml
```diff
@@ -38,7 +38,7 @@
     </check>
     <check level="error" class="org.scalastyle.scalariform.ParameterNumberChecker" enabled="true">
         <parameters>
-            <parameter name="maxParameters"><![CDATA[8]]></parameter>
+            <parameter name="maxParameters"><![CDATA[9]]></parameter>
```
tdroxler, line 41: https://github.com/alephium/explorer-backend/issues/455

tdroxler:
I added a small commit to update our java version in dockerfile, as [defined in the node](https://github.com/alephium/alephium/blob/86c2375e902815fe2739ddc441b416eaf140c69b/docker/release/Dockerfile.release#L1)

simerplaha (approved):

Issue #348, opened 2022-09-14
IO count report


simerplaha:
Working on [this](https://github.com/alephium/explorer-backend/issues/343#issuecomment-1243204866) issue I kinda needed a proper IO count check because from the code it seemed high.

Looks like we are executing around 1000 queries a second, which is high.

The graph below shows the number of queries executed by `DBRunner` per second, for the first 10 minutes (while also running a node locally).

If you'd like to inspect this graph, I've uploaded it [here](https://simerplaha.github.io/alephium-explorer-sandbox/).

- Green shows IO count when running `BlockFlowSyncService` only.
- Red show IO count for running all services.

![image](https://user-images.githubusercontent.com/1773953/190069295-dff8cb01-7efb-4c5f-ae42-b3ab2de07597.png)

Let me know if you'd like to run these IO count yourself, I will push some code so it's executable. 

Is this a valid issue?

## Concurrent overwrites

One thing of interest is [here](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L111), data fetched from multiple nodes is concurrently overwriting data created by one another other. I'm sure this will eventually lead to the correct database state but is this concurrent mutation by design/required?

simerplaha:
Maybe someone else should run this IO count locally? Incase I'm making some mistake, 1000 queries a second is a bit too high to be true.

polarker:
1000 queries per second, that's indeed way too high. Is it the number after syncing?

tdroxler:
>One thing of interest is [here](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/service/BlockFlowSyncService.scala#L111), data fetched from multiple nodes is concurrently overwriting data created by one another other 

Actually  we are not fetching same data. In our sharded blockchain, we can run a clique of multiple nodes and each node is responsible for a subset of chain indexes. In order to get all data we need to query every node.
the `nodeUris` is given on start when fetching the clique info.

simerplaha:
> Is it the number after syncing?

It's during sync. 

Node is here
[...]

PR #456, opened 2023-04-14
Stream for alph amount history

tdroxler:
Still a POC as I need to test and clean some stuff, but as I'll be out for a week it might be good to start reviewing, the global idea is here.

This PR was released as https://github.com/alephium/explorer-backend/releases/tag/v1.13.5-rc2

It might not be clear but the related issue is https://github.com/alephium/desktop-wallet/issues/592

I'm actually surprise by the good performances :sweat_smile:  when I remove the one year time restriction, it takes around 20/30s to get the all time data [for our biggest miner](https://explorer.alephium.org/addresses/1Gzp3Wc42fanrbAqknYmSaZLfC567bP3JWprDCiuJ8rbP) and its million of txs. 

Data will be cached on wallet front-end side.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
@@ -16,6 +16,9 @@
 
 package org.alephium.explorer.service
 
+import java.math.BigInteger
+import java.time.Instant
+
 import scala.collection.immutable.ArraySeq
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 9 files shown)


app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
```diff
@@ -238,6 +251,73 @@ object TransactionService extends TransactionService {
...
+    val timeranges = amountHistoryTimeRanges(from, to, intervalType)
+
+    Source(timeranges)
+      .mapAsync(paralellism) {
+        case (from, to) =>
+          run(sumAddressOutputs(address, from, to)).map(res => (res, from, to))
+      }
+      .mapAsync(paralellism) {
+        case (outs, from, to) =>
+          run(sumAddressInputs(address, from, to)).map(res => (outs, res, to))
```
simerplaha, line 271: Neat! Streams are cool 👍🏼. Parallel streams would add to the [total IO](https://github.com/alephium/explorer-backend/issues/348) I guess.

These queries seem to do sequential traversal instead of random searches, kinda wonder if parallelism would do well here. 

Surely there also other other back-pressure requirements I'm unaware of. This is cool! 👍🏼

tdroxler:
I will close and clean this branch before opening a new PR, a lot of changes occurs due to https://github.com/alephium/explorer-backend/pull/457

Issue #438, opened 2023-03-13
Replace akka-streams with RxJava


tdroxler:
Currently we use akka-streams only for the csv export. 
We should explore https://github.com/ReactiveX/RxJava which is  a very popular library + it has only a single dependency on [reactive-streams](https://github.com/reactive-streams/reactive-streams-jvm).
Also [vertx has an official api for it](https://vertx.io/docs/4.1.8/vertx-rx/java3/)

tdroxler:
I did successfully did it in this branch https://github.com/alephium/explorer-backend/tree/feature/java-rx

Performance is similar, I couldn't  see any difference.
jar file was reduced from 108M to 104M, so no big difference.

I think for now we won't move forward as it doesn't seems to bring a significant improvement.

PR #457, opened 2023-04-27
Feature/java rx

### LINKED ISSUE #401: Fix `AddressServerSpec` export transactions flaky test
all `/addresses/<address>/export-transactions/`  tests are flaky, mostly in CI.

I couldn't find the issue, but you can enforce the error by requesting calling a lot  the endpoint like in this [branch/commit](https://github.com/alephium/explorer-backend/compare/feature/replace-sttp-with-vertx-backend-test) .

I first thought that maybe we were not capable of handling lots of such call, but I did the following on a running the explorer backend : 
```
while [ true ]
do
  curl -s "localhost:9090/addresses/1GxJC7G5HNiDAeAw224PTjNzPFwcto7A9xmP1d8JGzqZg/export-transactions/csv?toTs=1668701628067&fromTs=1637362800000"
done
```
Where the address as the same range of txs than in the test. I let it run for several minutes without any issues.
I also ran: 
```
seq 1 100 | xargs -n1 -P10  curl -s "localhost:9090/addresses/1GxJC7G5HNiDAeAw224PTjNzPFwcto7A9xmP1d8JGzqZg/export-transactions/csv?toTs=1668701628067&fromTs=1637362800000"
```
Which query in parallel 100 times and it eventually completes.

As you can see in the commit linked above, I then tried to change our http client from `sttp` to `vertx`, thinking it could come from it, but it was still failing.

In order to merge https://github.com/alephium/explorer-backend/pull/389 I will reduce the number of call made during the tests.

The all point of this issue is to fix the problem and get back with a lot of calls.

See [this commit](https://github.com/alephium/explorer-backend/pull/389/commits/85bc7e97a09945427a4236b86ec5e6816c19e113)

tdroxler:
Resolves #401 and #438

I first thought that the above issue was a super low priority, but actually the FE team got hit by this it when testing [this PR](https://github.com/alephium/explorer-backend/pull/456) , I could then also reproduce locally, but it was a pain the find the source of the problem.

It seems there is an issue with `vertx-reactive-streams`, it doesn't handle well back pressure and in case of heavy load, it can just hangs a request forever, without any error.

To be able to link `akka-streams` and `vertx`, we need to pass through this `vertx-reactive-streams`, so it seems the right time to migrate to `rx-java` as there are builtin integrations in `vertx` for it. 

One of the hanging point was the perfs, that weren't that better, but actually with this following change it improved a lot: 

> Use Flowable.fromCompletionStage instead of fromFuture

`fromFuture` is actually performing some blocking operations and is not recommended.
This little change improve a lot the export of txs. With the former we were reaching 600k/s, up to 700k/s when jvm and db was hot.
With the second one, it reaches around 1400k/s when cold and up to 2700k/s when hot on my various tests.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
@@ -18,10 +18,10 @@ package org.alephium.explorer.service
 
 import scala.collection.immutable.ArraySeq
 import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+import scala.jdk.FutureConverters._
 
-import akka.NotUsed
-import akka.actor.ActorSystem
-import akka.stream.scaladsl._
+import io.reactivex.rxjava3.core.Flowable
 import io.vertx.core.buffer.Buffer
 import org.reactivestreams.Publisher
 import slick.basic.DatabaseConfig
@@ -32,7 +32,6 @@ import org.alephium.explorer.cache.TransactionCache
 import org.alephium.explorer.persistence.DBRunner._
 import org.alephium.explorer.persistence.dao.{MempoolDao, TransactionDao}
 import org.alephium.explorer.persistence.queries.TransactionQueries._
-import org.alephium.json.Json.write
 import org.alephium.protocol.model.{Address, TokenId, TransactionId}
 import org.alephium.util.{TimeStamp, U256}
 
@@ -108,17 +107,14 @@ trait TransactionService {
 
   def hasAddressMoreTxsThan(address: Address, from: TimeStamp, to: TimeStamp, threshold: Int)(
       implicit ec: ExecutionContext,
-      ac: ActorSystem,
       dc: DatabaseConfig[PostgresProfile]): Future[Boolean]
 
   def exportTransactionsByAddress(address: Address,
                                   fromTime: TimeStamp,
                                   toTime: TimeStamp,
-                                  exportType: ExportType,
                                   batchSize: Int,
                                   paralellism: Int)(
       implicit ec: ExecutionContext,
-      ac: ActorSystem,
       dc: DatabaseConfig[PostgresProfile]): Publisher[Buffer]
 }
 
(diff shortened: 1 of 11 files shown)


app/src/main/scala/org/alephium/explorer/service/TransactionService.scala
```diff
@@ -244,39 +235,34 @@ object TransactionService extends TransactionService {
...
-      .grouped(batchSize)
-      .mapAsync(paralellism) { hashes =>
-        run(getTransactionsNoJoin(ArraySeq.from(hashes)))
-      }
+      .buffer(batchSize)
+      .concatMapEager(({ hashes =>
+        Flowable.fromCompletionStage(
+          run(getTransactionsNoJoin(ArraySeq.from(hashes.asScala))).asJava
+        )
+      }), paralellism, 1)
```
simerplaha, line 246: So that the `Stream` handles back-pressure more effectively, we could consider replacing the mid-stream query execution using `run(getTransactionsNoJoin ...`) with a `Publisher` that is returned without executing a query? For example:

```scala
Flowable
  .fromPublisher(stream(streamTxByAddressQR(address, from, to)))
  .buffer(batchSize)
  .flatMap(hashes => stream(getTransactionsNoJoin_Streamed(ArraySeq.from(hashes.asScala))))
```

`getTransactionsNoJoin` would need to be stream-fied for this.

tdroxler: Totally agree, I'll check if it's possible to convert the query into a stream....Slick can be a pain sometimes.....

tdroxler: I had a quick look and it will be more complex than expected, because [here](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala#L335-L341) we need to get all inputs/outputs and it's a non-streaming action. Then Slick doesn't love it and it seems we can't go back to a streaming.  But I'll open an issue for that, because it would be definitively cool to have a full reactive pipe from the http request up to the response.

simerplaha (approved):
👍🏼

polarker (approved):
I don't have time to check out rxjava, but the numbers looks convincing.

PR #462, opened 2023-05-08
Use parametric queries instead of spliced values

tdroxler:
related to https://github.com/alephium/explorer-backend/issues/426


simerplaha (approved):
👍🏼

PR #464, opened 2023-05-10
Clean all our functions name

tdroxler:
During our long phase of optimizing our queries, we were keeping old functions and introducing new ones in parallel, to test they were returning the same result.

We were always adding some `SQL` and/or `NoJoin` to be explicit.

Now that it's stable since a long time, we already removed the old function, but didn't adapts the naming.

This PR clean all our names, removing all the `SQL/NoJoin` that made the code dirty and confusing.

I also removed few functions that were not used anymore.

Sorry for the boring PR to review, but it should be only naming things.


tdroxler:
FYI I ran the explorer FE locally, everything works perfectly fine.

simerplaha (approved):
All nice and clean now 👍🏼.

Big PR for a single commit though. Would've been easier to review if every renaming/removal and its replacement were in their own individual commits. But it's ok 👍🏼 

Ran some tests, looks good.

tdroxler:
> Big PR for a single commit though. Would've been easier to review if every renaming/removal and its replacement were in their own individual commits. But it's ok

sorry, the git extremist in me was too lazy this time :sweat_smile:

PR #466, opened 2023-05-16
Clean `Main`

tdroxler:
We were previously having different boot up for `Read...` and `WriteOnly`, because the first ones needed an `ActorSystem`. Now that we use RxJava, we can unify everything.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/Main.scala
+++ b/app/src/main/scala/org/alephium/explorer/Main.scala
@@ -54,67 +54,16 @@ class BootUp extends StrictLogging {
   implicit val databaseConfig: DatabaseConfig[PostgresProfile] =
     DatabaseConfig.forConfig[PostgresProfile]("db", typesafeConfig)
 
-  def init(): Unit =
-    config.bootMode match {
-      case mode: BootMode.Readable =>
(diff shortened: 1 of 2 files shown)


app/src/main/scala/org/alephium/explorer/Main.scala
```diff
@@ -54,67 +54,16 @@ class BootUp extends StrictLogging {
...
-    explorer
-      .start()
-      .onComplete {
-        case Success(_) => ()
-        case Failure(error) =>
-          logger.error("Fatal error during initialization", error)
-          explorer.stop().failed foreach { error =>
-            logger.error("Failed to stop explorer", error)
-          }
-      }
```
simerplaha, line 100: I didn't realise that on master `explorer.start()` is called twice.

simerplaha (approved):
🚀

simerplaha (approved):
🚀

app/src/main/scala/org/alephium/explorer/Main.scala
tdroxler: ahah yeah my fault on previous refactor. Hopefully our `Service` can't be started twice.

Issue #86, opened 2021-11-19
Improve `TokenSupplyService`


tdroxler:
* We could probably cache the genesis blocks as we always query the same
* `unspentTokens` function can be very slow with millions of utxos
* The syncPeriod should be adaptive, when we know computed all supply, we can wait for 1 all day.

tdroxler:
We changed how we count the token supply recently, not sure the description is still relevant, but let's keep the issue so we check.

tdroxler:
We could now probably get rid of the `JOIN` in our token supply queries, as we are now updating the inputs with the output ref info. But we need to be sure we only compute token supply when the inputs are updated.

tdroxler:
`spent_finalized` isn't inserted either because when we insert an output it can't be spent yet, but while writing I realize we should insert both value anyway even if it's None most of the time, we don't know in the future if suddenly we'll insert spent outputs. Nice catch

simerplaha:
I think the test-cases couldn't pick this up because the generator [`outputEntityGen`](https://github.com/alephium/explorer-backend/blob/token-supply-improvement/app/src/test/scala/org/alephium/explorer/Generators.scala#L532C11-L533) always default these fields to `None`.

tdroxler:
> I think the test-cases couldn't pick this up because the generator [`outputEntityGen`](https://github.com/alephium/explorer-backend/blob/token-supply-improvement/app/src/test/scala/org/alephium/explorer/Generators.scala#L532C11-L533) always default these fields to `None`.

good point, I'll add a test for that case

tdroxler:
> I think the test-cases couldn't pick this up because the generator [`outputEntityGen`](https://github.com/alephium/explorer-backend/blob/token-supply-improvement/app/src/test/scala/org/alephium/explorer/Generators.scala#L532C11-L533) always default these fields to `None`.

done in [this commit ](https://github.com/alephium/explorer-backend/pull/470/commits/bcf47f2626b740c3540c1f5685f97a086cc196d4) You'll see it in [this PR](https://github.com/alephium/explorer-backend/pull/470/commits)

tdroxler:
done in https://github.com/alephium/explorer-backend/pull/470

PR #469, opened 2023-05-26
Spent timestamp

tdroxler:
As explain in [this issue](https://github.com/alephium/explorer-backend/issues/86)

A Release will need to be done and run over our network before the next one.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
@@ -37,7 +37,6 @@ import org.alephium.explorer.persistence.DBRunner._
 import org.alephium.explorer.persistence.model.TokenSupplyEntity
 import org.alephium.explorer.persistence.schema._
 import org.alephium.explorer.persistence.schema.CustomGetResult._
-import org.alephium.explorer.persistence.schema.CustomJdbcTypes._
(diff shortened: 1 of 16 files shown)


app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
```diff
@@ -223,12 +222,15 @@ case object TokenSupplyService extends TokenSupplyService with StrictLogging {
...
-              .result
-              .headOption
+            sql"""
+              SELECT block_timestamp
+              FROM block_headers
+              WHERE main_chain = true
+              AND chain_from = $from
+              AND chain_to = $to
+              ORDER BY block_timestamp DESC
+              LIMIT 1
```
simerplaha, line 232: The typed query being replaced is not querying for `main_chain = true`. Is the `main_chain` check in SQL query something that was missing from the typed query?

tdroxler: Good point, I think it actually doesn't change much with or without it, token supply is computed daily at midnight, the latest  block is more like a time reference to see when we change day. 
As it's the same it's maybe cleaner to have the `main_chain = true`?

simerplaha (approved):
👍🏼

polarker (approved):
LGTM

tdroxler:
I'll make a release on Monday and deploy it on our various env to let the new field spread out.

PR #470, opened 2023-05-26
Token supply improvement

tdroxler:
Following of [this PR](https://github.com/alephium/explorer-backend/pull/469)


app/src/main/scala/org/alephium/explorer/service/TokenSupplyService.scala
```diff
@@ -80,15 +79,20 @@ case object TokenSupplyService extends TokenSupplyService with StrictLogging {
...
+                                     scheduler: Scheduler): Future[Unit] = {
+    //Sync once on start to make sure we are up to date and then sync once a day at the given time.
+    syncOnce().map { _ =>
+      scheduler.scheduleDailyAt(
+        taskId = TokenSupplyService.productPrefix,
+        at = ZonedDateTime
+          .ofInstant(Instant.EPOCH, ZoneOffset.UTC)
+          .plusSeconds(scheduleTime.toSecondOfDay().toLong)
+      )(syncOnce())
+    }
```
simerplaha, line 94: Question: Could we run into a situation where a couple of days pass by without us knowing that this service is not running?

If a `synOnce()` run fails for whatever reason, the `Scheduler` simply logs the failure and does not reschedule for the next day. Right?

So, if a failure does occur and we wanna the cause, is our option only to search for the `Exception` in that day's log file? Or does the magic of devops notify us or a health check endpoint?

tdroxler: if `syncOnce` is failing we don't only log the error, we also[ report the failure to the execution context](https://github.com/alephium/explorer-backend/blob/8840602ae10345936efe614116136fcadf528b3d/app/src/main/scala/org/alephium/explorer/util/Scheduler.scala#L136-L138) . The app is then shutdown.
I just gave a try and manually failed inside the `syncOnce`, the app stops well.

simerplaha (approved):
👍🏼

polarker (approved):
LGTM

Issue #468, opened 2023-05-22
Clean and organize all our generators file


tdroxler:
we have multiple file with all our scala check generators
 * Generators
 * GenCoreApi
 * GenDBModel
 * GenApiModel
 * GenCoreUtil

it's sometimes hard to find the `Gen` we want, also `Generators` has a lot of `Gen` that could be moved to some more specific file.

PR #471, opened 2023-05-26
Clean generators

tdroxler:
Resolves https://github.com/alephium/explorer-backend/issues/468
  here are the interseting comments I made in some commits: 

* 8d25dd17061a8a87ff977bc70a1ee1add66700b6

    Reorganize some of `Gen`
    
    The most noticeable change is the move of `BlockHash`, `TransactionId`
    generators into `GenCoreProtocol`.
    
    They were in `GenApiModel` but it was confusing as we use the types from
    `org.alephium.protocol`
    
    Some `*Entity` model were also moved from `Generators` into `GenDBModel`

* 50054a13821f473ccdb7f8d1298f5adaf3e9b2e0

    Introduce `ConfigDefaults` for test
    
    Having the group setting implicit value available is super useful, but
    until now it was in the `Generators` object, with the incoming changes
    it was creating some dependencies loop problem.
    
    Having it in a proper and isolated object make it easier to fine grain
    our tests.

* 5fc9c276923e2cdd80c96b68cf3ab23b532094bf

    Unify block hash generator


simerplaha (approved):
Neat! It was getting a bit inconvenient to find a generator 👍🏼. I also thought you'd revert back to - "everything in `Generator.scala`" :), but this is nice 👍🏼👍🏼.

Issue #472, opened 2023-05-26
Use more models comming from the node.


tdroxler:
We have some duplicates, like `GroupIndex`, it could make sense to use the one comming from the node directly, trying to reduce the code base to maintain.

PR #473, opened 2023-06-02
Models from node

tdroxler:
A long and boring PR to review :sweat_smile: 
But commits should be well splitting and it's mostly about renaming the imports.

The goal is to remove the duplicates we have between the full node api and clean few other stuff.

Resolves: https://github.com/alephium/explorer-backend/issues/472

image

 Less lines of code to maintain :muscle:


simerplaha (approved):
💪🏼

simerplaha (approved):
💪🏼

PR #474, opened 2023-06-06
Add AppState `migrationsVersion` and `lastFinalizedInputTime` to `ExplorerInfo`

tdroxler:
The data in database is serialized to bytestring and it's complicated to get back the original value.


simerplaha (approved):
👍🏼

Issue #353, opened 2022-09-15
Sync using full node's web socket


tdroxler:
Once the explorer-backend is up-to-date with the sync, we could maybe use the [websocket](https://github.com/alephium/alephium/blob/master/app/src/main/scala/org/alephium/app/WebSocketServer.scala) to sync new blocks to avoid polling.

Very low priority as it can bring more complexity and anyway our polling of 5 seconds almost got a new block every time.  Bu it might be good to explore this one day.

tdroxler:
Latest branch is here https://github.com/alephium/explorer-backend/tree/websockets

PR #478, opened 2023-06-23
Web socket syncing

tdroxler:
Resolves: https://github.com/alephium/explorer-backend/issues/353 

After thousands of tests and adjustment I think we can start the discussion, it still needs some unit tests, but I'm not sure yet how easy it will be to test the websockets.

The PR needs https://github.com/alephium/dev-alephium/pull/827 from the node, because we need the `events` to match the data we receive from rest calls.

The strategy is:
* Sync with `BlockFlowSyncService` as usual
* Once we are up-to-date: latest block is less than 30s old we move to
  websocket syncing
* If websocket messages are late: > 30s old, we close the websocket and
  move back to http syncing. This can happen in case of network issue,
  this is safer as we could suddenly receive lots of message through the
  websocket and the DB can't follow (happen once in my tests)
* If the websocket close for anyreason we also move back to http
  syncing.

The main idea is to always rely and fall back on our `BlockFlowSyncService` which is well tested in production.

I tried various edge cases, like cutting my network etc.


simerplaha:
Neat! We have WebSockets! Is the way to test this is to do a full sync and try to manually terminate the server to test how it behaves?

Tricky one to test. I'm letting it sync overnight and kinda manually causing abrupt server shutdowns to see how the connection & sync behaves with debug statements which kinda sometimes corrupts my RocksDB instance and I have to start over.  

Surely there is an easier way to test. Any suggestion?

> it still needs some unit tests, but I'm not sure yet how easy it will be to test the websockets.

Yep, tricky one indeed.

simerplaha:
Oh it's not a bug. It's me intentionally causing abrupt/force shutdown to see how the connection behaves with `println` debug statement. Node and explorer behave as expected. Not sure how else to test it. Was trying to convey that testing is tricky, as you've pointed out.

tdroxler:
It changes a lot and I will create a new PR with a new branch.

PR #502, opened 2023-12-04
Fix fetch of metadata

tdroxler:
We were using a deprecated way of fetching nft metadata, we now follow what's inside our web3 package:

https://github.com/alephium/alephium-web3/blob/72ae2e0ef8c0274ffb3e0cedb33f609585b5757a/packages/web3/src/api/node-provider.ts#L136-L180

I tested it both in mainnet and testnet, everything seems fine. Of course on testnet we have some nft that are deprecated so we can't get their metadata, but that's expected.

@h0ngcha0 I put you as reviewer as you wrote the web3 part, you confirm that now `collectionId` and `index` are in the same `CallContractSucceeded`?

First elem is the collection_id and second is the index.


h0ngcha0 (approved):
LGTM :+1:

PR #503, opened 2023-12-15
Price service

tdroxler:
Here is a first version, I still want to add proper types for our currencies and coins, so we could expose them in the `open-api.yaml` for the FE team, but could be nice to have a first review especially for the retry things.

The retry mechanism happen [here](https://github.com/alephium/explorer-backend/compare/price-service?expand=1#diff-76f61b3c4df860f87d76b60687f32eee60268e2d2592a5c2933cba8e702b93eaR263-R284).

The idea is quite simple, if the request return a `TooManyRequest` we retry after X minutes in the same future, so the cache only get update at the end when we finally get a result.

I used our `AsyncReloadCache`, that always returns the latest data, even if expired, so it's fast and we don't have a risk to wait for coingecko to respond.

The only draw back is on start, we need to load data and if someone is requesting right on start, he might get an error. But I think a user could live 2 minutes without a price value.


mvaivre:
Thanks! Testing it now. What about making the endpoint params case insensitive? For now, if we passe `ALPH` instead of `alph` it will fail silently. Same with currencies.

tdroxler:
Recap with the latest changes discussed with the FE team.
* We now use only `TokenId` instead of name or symbol, as it's the better and only real source of truth.
*  We change the format type for the chart price, we use a proper type rather than an array of `(Long, Double)`, This create better TS types and on FE side, they don't have to do some casting.

app/src/main/resources/explorer-backend-openapi.json
```diff
@@ -6219,13 +6219,12 @@
...
-                  "type": "string",
-                  "format": "32-byte-hash"
+                  "type": "string"
                 }
               },
               "example": [
-                "2d11fd6c12435ffb07aaed4d190a505b621b927a5f6e51b61ce0ebe186397bdd",
-                "bd165d20bd063c7a023d22232a1e75bf46e904067f92b49323fe89fa0fd586bf"
+                "ALPH",
+                "USDT"
```
polarker, line 6227: Maybe we should provide the complete list? Or make it an enum?

Otherwise, people might not know what to provide

Issue #506, opened 2024-01-18
Use deltas to show amount history


tdroxler:
We made good progress [in this PR](https://github.com/alephium/explorer-backend/pull/498) to improve the amount history charts.

But to do even better we need to change the format and use deltas instead of actual amount for a given timestamp.

It will implies some changes on the FE side.

Instead of having a fix amount per timestamp, we will work with deltas.
If a user request from January 7 to January 9, we will show and compute ONLY:
Jan 7: +4 ALPH
Jan 8: -6 ALPH
Jan 9: +1 ALPH
If the user want to have it's full history balance, as the FE team will do for wallets, they will need first to fetch and cache the deltas from the beginning, then we re-opening the wallet, only the missing data will be fethched.

In the example above we see that if we do the math it's : -1 ALPH, but because we could have Jan 1: +100 ALPH.

I made a [POC here](https://github.com/alephium/explorer-backend/tree/amount-history-improvement) The perfromance are much better

PR #508, opened 2024-01-30
Add new endpoint `/addresses/<address>/amount-history-deltas`

tdroxler:
Resolves: https://github.com/alephium/explorer-backend/issues/506 the new approach is described in that issue.

I currently keep the previous endpoint as FE will need to adapt first. Here are some results of the benchmarks:

image
image
image


tdroxler:
I'm closing as we discussed with the team and we will rather download the transactions history and get the token amount history from there

PR #511, opened 2024-02-12
Open api improvement

tdroxler:
As discussed @polarker 

There was already the max items info, but it was a bit hidden in the schema tab
image
So I added a description 
image


polarker (approved):
LGTM

PR #519, opened 2024-03-04
Add more config for yearly data and default to 366 to handle leap years

tdroxler:
From the front end, if you query for march 05 2023 to march 05 2024, then you have 366 days as it's a leap year. 

I think we can safely put 366 as default max value, one day won't change much and at least we are fine with leap years.

I added more config value so now every thing can be change from the config

cc @mvaivre

### DIFF
--- a/app/src/main/resources/application.conf
+++ b/app/src/main/resources/application.conf
@@ -50,19 +50,21 @@ explorer {
         amount-history = {
             hourly = 7 days
             hourly = ${?EXPLORER_AMOUNT_HISTORY_MAX_TIME_INTERVAL_HOURLY}
-            daily  = 365 days
+            daily  = 366 days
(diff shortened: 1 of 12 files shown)


app/src/main/resources/application.conf
```diff
@@ -87,7 +89,9 @@ explorer {
             "try",
             "cad"
         ]
-        coingecko-uri = "https://api.coingecko.com/api/v3"
+        coingecko-uri = "https://api.coingecko.com/api/v3",
```
simerplaha, line 92: Is the comma a typo?

simerplaha (approved):
Just one minor question.

app/src/main/resources/application.conf
tdroxler: more like a JSON reflex ^^ nice catch

PR #526, opened 2024-03-21
Update to node `2.11.0`

tdroxler:
The new release after the merge will need a full re-sync to handle the new event  change.


polarker:
LGTM. Let's make a new release so that @sven-hash could resync his full node and explorer backend!

polarker (approved):

PR #529, opened 2024-03-28
Add `main_chain` to contracts and events

tdroxler:


tdroxler:
We could update and merge this PR the next time we need a full resync

PR #545, opened 2024-05-31
Add Welcome message when explorer-backend is succesfully started

tdroxler:
The current info logs were not very clear if the explorer-backend was really started or not, so users were confused.

In this PR when the explorer-backend is successfully started, we then logs useful info. Here are the 3 possible cases based on the mode.
image
image
image


h0ngcha0 (approved):
LGTM :+1:

PR #546, opened 2024-06-04
Expose currencies and tokens with price in openapi

### LINKED ISSUE #512: Provide supported market tokens through an openapi enum
currently we have the list in our config:https://github.com/alephium/explorer-backend/blob/5780248bde5b5323ff9846af88be8ddb3b41a380/app/src/main/resources/application.conf#L68-L90 

This could be useful if we were also exposing that list through the openapi.

cc @mvaivre

tdroxler:
Resolves #512
 
The generated TS types are then: 
```ts
export enum TokensWithPrice {
  WETH = 'WETH',
  ALPH = 'ALPH',
  USDT = 'USDT',
  AYIN = 'AYIN',
  DAI = 'DAI',
  USDC = 'USDC',
  WBTC = 'WBTC'
}

export enum Currencies {
  Btc = 'btc',
  Usd = 'usd',
  Eur = 'eur',
  Chf = 'chf',
  Gbp = 'gbp',
  Idr = 'idr',
  Vnd = 'vnd',
  Rub = 'rub',
  Try = 'try',
  Cad = 'cad',
  Aud = 'aud'
}
```

### DIFF
--- a/app/src/main/resources/application.conf
+++ b/app/src/main/resources/application.conf
@@ -79,6 +79,7 @@ explorer {
         }
         currencies = [
             "btc",
+            "eth",
             "usd",
             "eur",
             "chf",
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -8338,6 +8338,34 @@
         "enum": [
           80
         ]
+      },
+      "TokensWithPrice": {
+        "type": "string",
+        "enum": [
+          "WETH",
+          "ALPH",
+          "USDT",
+          "AYIN",
+          "DAI",
+          "USDC",
+          "WBTC"
+        ]
+      },
+      "Currencies": {
+        "type": "string",
+        "enum": [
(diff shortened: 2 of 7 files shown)


nop33 (approved):
Not sure what I can review here, but it's cool to see that BTC is an available currency! That's something our community asked for: https://github.com/alephium/alephium-frontend/issues/98

Maybe we can also have ETH? Do we want that?

tdroxler:
> Not sure what I can review here

Was mostly so you're aware of the new types

>Maybe we can also have ETH? Do we want that?

It's available on [coingecko](https://api.coingecko.com/api/v3/simple/price?ids=alephium&vs_currencies=eth) so fine for me, any thought @mvaivre ?

nop33:
The person who requested this mentioned:

> A BTC or ETH valuation instead of fiat would be a nice touch. Especially ETH since we are bridging to it, but BTC also for more adoption. A lot of people don't think in terms of fiat anymore

PR #548, opened 2024-06-12
Quickfix Rhone hashrate

tdroxler:
I first started with those value in the config file, but passing this everywhere until we need it is actually trickier that I thought, especially in tests.

So I propose this quickfix first and so we could quickly deploy and we clean in a second PR.


Lbqds (approved):
LGTM 👍

PR #550, opened 2024-06-13
Add missing data

tdroxler:
Big PR, but don't freak out, lot's of function moving and adding fields to tables.

There are 3 topics in this PR:

### Remove the `block_deps` table

when first started to implement the explorer-backend, the idea to store the `deps: ArraySeq[BlockHash]` was to use the traditional sql way, with a separate table with one row for each element of the array, so we could make some query and join on it. 
It actually happens that we use that table just to get the deps when getting a block, so the deps could simply be stored as a serialized data in the block table. First it's more efficient to fetch a block, no need to query another table, but morever...... the block_deps table was actually our biggest table :scream: :scream: :scream: 

image

I'm still syncing from scratch a new db with this PR to see the difference, but for sure the serialized data in the block table should be much less than the block_deps table.

### Rework model of `/block/<hash>` endpoint

In order to make  [block pages](https://explorer.alephium.org/blocks/00000000000248d5bddb6d3e42b693252c7bf88023717952cc6d1a0caec23b72) more efficient, the `/block/<hash>` was returning the `BlockEntryLite` version and the transactions were then fetched with `/block/<hash>/transactions` which has pagination.
I think it's good to keep that way, as querying transactions is an heavy task.
In this PR the endpoint now returns `BlockEntry`. I added all missing informations, like `ghostUncles`, `txsHash`, `nonce`  etc, except the `transactions` so they are still fetched with the dedicated paginated endpoint. This is all backward compatible as it's just adding fields, no one is removed.

So if a user want some info on a block, but doesn't care about txs, he can efficiently query all information with that endpoint.

### Make sure we always store all block data

That's the part that bring a bit of noise in this PR, as I added every missing data in all our models to make sure we have everything in DB. I then created a [test](https://github.com/alephium/explorer-backend/blob/58145ec5e0f7f25a0e47739bf090a52731a233e0/app/src/test/scala/org/alephium/explorer/BlockModelConversionSpec.scala#L30-L38) that makes sure we can convert the `Block` coming from the full-node, into our DB models, then to the explorer-backend API models and from those one back to the full node models. That's all the `toProtocol` and `toApi` functions

So now we should be able to provide any data requested by user.

Take your time, I'm anyway re-syncing locally to make sure everything is correct.


polarker:
@h0ngcha0 @Lbqds Please help me review this with priority!

app/src/main/scala/org/alephium/explorer/api/model/Transaction.scala
```diff
@@ -59,6 +66,30 @@ final case class Transaction(
       .getOrElse("")
     s"${hash.toHexString},${blockHash.toHexString},${timestamp.millis},$dateTime,$fromAddressesStr,$toAddresses,$amount,$amountHint\n"
   }
+
+  @SuppressWarnings(Array("org.wartremover.warts.TripleQuestionMark"))
```
Lbqds: [Nit]: It looks like we don't need this

app/src/main/scala/org/alephium/explorer/persistence/model/BlockHeader.scala
```diff
@@ -41,11 +42,38 @@ final case class BlockHeader(
...
     target: ByteString,
     hashrate: BigInteger,
-    parent: Option[BlockHash]
+    parent: Option[BlockHash],
+    ghostUncles: Option[ArraySeq[GhostUncle]]
 ) {
   def toApi(deps: ArraySeq[BlockHash], transactions: ArraySeq[Transaction]): BlockEntry =
     BlockEntry(hash, timestamp, chainFrom, chainTo, height, deps, transactions, mainChain, hashrate)
 
+  @SuppressWarnings(Array("org.wartremover.warts.TripleQuestionMark"))
```
Lbqds: Same here.

app/src/test/scala/org/alephium/explorer/api/model/BlockEntryTest.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.collection.immutable.ArraySeq
+
+import akka.util.ByteString
+
+import org.alephium.explorer.service.FlowEntity
+import org.alephium.protocol.Hash
+import org.alephium.protocol.model.{BlockHash, GroupIndex}
+import org.alephium.util.TimeStamp
+
+final case class BlockEntryTest(
```
Lbqds, line 30: Question: `BlockEntryTest` seems the same as `BlockEntry`. Can we directly use `BlockEntry` for testing?

Lbqds (approved):
LGTM 👍. A question unrelated to this PR: it seems that explorer-backend sync blocks using the `/blockflow/blocks` endpoint, but this endpoint can only sync mainchain blocks. How does the explorer-backend sync uncle blocks?

app/src/main/scala/org/alephium/explorer/api/model/Transaction.scala
tdroxler: :+1:  development leftover

app/src/test/scala/org/alephium/explorer/api/model/BlockEntryTest.scala
tdroxler: `BlockEntryTest` has `transactions`, while `BlockEntry` hasn't to serve the api separately in another paginated endpoint

tdroxler:
>it seems that explorer-backend sync blocks using the /blockflow/blocks endpoint, but this endpoint can only sync mainchain blocks. How does the explorer-backend sync uncle blocks? 

Are uncles blocks guaranteed to be stored on every full-node? if yes we could add a mechanism to download them individually with [/blockflow/main-chain-block-by-ghost-uncle/{ghost_uncle_hash}](https://node.mainnet.alephium.org/docs/#/Blockflow/getBlockflowMain-chain-block-by-ghost-uncleGhost_uncle_hash)

app/src/main/scala/org/alephium/explorer/persistence/schema/CustomGetResult.scala
```diff
@@ -195,6 +208,13 @@ object CustomGetResult {
   implicit val optionInterfaceIdGetResult: GetResult[Option[InterfaceIdEntity]] =
     (result: PositionedResult) => result.nextStringOption().map(InterfaceIdEntity.from)
 
+  implicit val ghostUnclesGetResult: GetResult[ArraySeq[GhostUncle]] =
```
h0ngcha0, line 218: [Nit] Seems that this is not needed.

h0ngcha0 (commented):
LGTM :+1: 
Only minor comments

We would need a complete resync for the existing explorer, right?

tdroxler:
>We would need a complete resync for the existing explorer, right?

yes

tdroxler:
@polarker when we insert a block, we check if there is a list of uncles and for each hash of that list we fetch the uncle block and insert it in DB.  
At that moment we could flag all those blocks with the `isUncle` flag, maybe even with the block hash that defined that uncle list? or it's more complicated than that?

polarker:
> @polarker when we insert a block, we check if there is a list of uncles and for each hash of that list we fetch the uncle block and insert it in DB. At that moment we could flag all those blocks with the `isUncle` flag, maybe even with the block hash that defined that uncle list? or it's more complicated than that?

`isUncle` is as complicated as `isMainChain`. An uncle block might become a mainchain block after a reorg.

A block can have the following statuses, which can change due to reorgs:
1. On the mainchain
2. As an uncle of a mainchain block
3. Orphan block: neither on the mainchain nor as an uncle

We could handle uncle blocks and make them aware of chain reorgs in two ways:

1. **On-Demand Query**: 
   - Don't handle the status preemptively and query the uncle status on demand.
   - In the front-end, when querying for a block and finding that it's not on the mainchain, we then try to query the mainchain block that contains it as an uncle.
   - If such a mainchain block is found, then it's an uncle block. Otherwise, we treat it as an orphan.

2. **Background Service**:
   - Have a background service that regularly updates the uncle status for new blocks that are finalized with enough confirmations.
   - This service will keep track of the status changes due to reorgs and ensure the uncle status is up-to-date.
   
  We could take approach 1 for now and switch to approach 2 later on.

app/src/main/scala/org/alephium/explorer/api/model/Output.scala
```diff
@@ -60,11 +63,79 @@ final case class ContractOutput(
...
+          )
+        )
+      case (_: ContractOutput, contractAddress: Address.Contract) =>
+        Some(
+          org.alephium.api.model.ContractOutput(
+            output.hint,
+            output.key,
+            org.alephium.api.model.Amount(output.attoAlphAmount),
+            contractAddress,
+            tokens = output.tokens
```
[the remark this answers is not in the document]
tdroxler: here we are converting to the[ full node API models](https://github.com/alephium/alephium/blob/6d5648ff43fa870afdf90886ec23492cb31de8de/api/src/main/scala/org/alephium/api/model/Output.scala#L31) which isn't an `Option`

PR #554, opened 2024-06-18
Amount history openapi

tdroxler:
Fixes: https://github.com/alephium/alephium-frontend/pull/680

the generated TS api is then: 
image

the data in array is actually `[ timestamp, string]`, but we can't represent such thing, so I checked with FE team and using string is all good.


mvaivre (approved):
Thanks! 👍

PR #555, opened 2024-06-25
Validate symbol and currency in market endpoints

tdroxler:
This avoid returning a `5xx` when someone query an unknown symbol or currency


h0ngcha0 (approved):
LGTM :+1:

PR #559, opened 2024-08-13
Fix concurrency issue in `Finalizer service`

tdroxler:
This PR addresses a concurrency issue in the `FinalizerService` that occurs during the finalization process while syncing  from scratch.

### Problem Description

When finalizing, we check the latest inserted input and compare its timestamp with the finalization threshold (`~(now - 1 hour)`). Inputs older than this threshold are finalized, and the app_state table is updated to reflect the finalization up to this point.

### Normal Operation:

In a fully synced explorer-backend, the latest input is typically newer than the finalization timestamp, allowing the process to complete without issue.

### Concurrency Issue

However, when syncing from scratch, we encounter a concurrency issue. Blocks are fetched in timestamp batches from the node, with 16 lists of blocks (one per chain) processed in parallel.

The problem arises when a block from one chain (e.g., (0,0)) that is newer is inserted before a block from another chain (e.g., (1,1)), which hasn't been inserted yet. If the FinalizerService triggers during this time, it finalizes based on the newer block's timestamp, marking it as the latest finalized point. Subsequently, when the older block of (1,1) is inserted, it is never finalized.

### Impact

This issue affects the computation of the circulating supply value only, as it relies specifically on the `spent_timestamp` field.

But with all my tests before my holydays on mainnet data seems to show that there's only an error of `93.61 ALPH` in the circulating supply, not much compared to the current `85.5M ALPH` supply.

The error was found thx to https://backend-v200.mainnet.alephium.org/infos which was synced from scratch and suffer a lot of this issue, especially now with Rhone and the increase number of blocks.

I'm currently syncing my local explorer-backend with this branch, I'll keep you updated with the final result once it's finished and I won't merged until I could check it.

Fortunately, it does not impact the balance computation of an address, as the `spent_timestamp` is used only to speed up queries.

### Solution

When checking the latest inputs, we take the 16 latest of each chain and
grab the oldest among them.

We also need to apply some migrations to fix the unfinalized outputs as well as resetting the supply table.

The supply will slowly be recomputed and eventually be up to date.

This won't impact https://explorer.alephium.org/ as it's still pointing to 1.19, the migration will be done on https://backend-v200.mainnet.alephium.org and once it's done, a snapshot will be created before we make 2.x the official latest release.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
@@ -31,7 +31,7 @@ import org.alephium.explorer.persistence.schema.CustomGetResult._
 @SuppressWarnings(Array("org.wartremover.warts.AnyVal"))
 object Migrations extends StrictLogging {
 
-  val latestVersion: MigrationVersion = MigrationVersion(1)
+  val latestVersion: MigrationVersion = MigrationVersion(3)
 
   def migration1(implicit ec: ExecutionContext): DBActionAll[Unit] = {
     // We retrigger the download of fungible and non-fungible tokens' metadata that have sub-category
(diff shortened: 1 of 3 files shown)


h0ngcha0 (approved):
LGTM :+1:

Issue #565, opened 2024-09-23
JVM Metrics


h0ngcha0:
Seems that the metrics endpoint doesn't return JVM metrics even though it is registered. It might have to do with how metrics are exposed [here](https://github.com/alephium/explorer-backend/blob/master/app/src/main/scala/org/alephium/explorer/Metrics.scala#L32-L57). Full node exposes the metrics endpoint using [CollectorRegistry](https://github.com/alephium/alephium/blob/master/app/src/main/scala/org/alephium/app/EndpointsLogic.scala#L840-L853).

PR #566, opened 2024-09-24
Jvm metrics

tdroxler:
Resolves: https://github.com/alephium/explorer-backend/issues/565

They are actually two different registries and `DefaultExports.initialize`
write to `CollectorRegistry` while tapir and HikariCP are writting to
`PrometheusRegistry`.

Only the second one was exposed through the metrics endpoint.


h0ngcha0 (approved):
LGTM, thanks :+1:

PR #567, opened 2024-09-26
Latest tx info

tdroxler:
* Reduce max page limit to 20 and default to 10
* Add /addresses/<address>/latest-transaction endpoint

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
@@ -97,6 +97,12 @@ trait AddressesEndpoints extends BaseEndpoint with QueryParams {
       .out(jsonBody[Int])
       .description("Get total transactions of a given address")
 
+  val getLatestTransactionInfo: BaseEndpoint[Address, TransactionInfo] =
+    addressesEndpoint.get
+      .in("latest-transaction")
+      .out(jsonBody[TransactionInfo])
+      .description("Get latest transaction information of a given address")
+
   val addressMempoolTransactions: BaseEndpoint[Address, ArraySeq[MempoolTransaction]] =
     addressesEndpoint.get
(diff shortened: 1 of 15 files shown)


h0ngcha0 (approved):
LGTM :+1:

app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
```diff
@@ -113,7 +113,7 @@ trait AddressesEndpoints extends BaseEndpoint with QueryParams {
   val listAddressTokens: BaseEndpoint[(Address, Pagination), ArraySeq[TokenId]] =
     addressesTokensEndpoint.get
       .out(jsonBody[ArraySeq[TokenId]])
-      .in(paginator(defaultLimit = 100))
+      .in(paginator(limit = 100))
```
polarker, line 122: Does it mean the default limit is 100 for tokens?

polarker (approved):
LGTM

app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
tdroxler: yes, this endpoint is just returning `TokenId` stored in one table, no heavy query like tx. But we could reduce it too if you prefer.

polarker: maybe run some benchmark on this to see the latency

tdroxler: Reading 100 or 10 row from DB is equivalent. 
image
image

polarker: Interesting. Then let's keep it unchanged for now

PR #568, opened 2024-10-01
Improve post addresses transactions

tdroxler:
`POST /addresses/transactions` was returning duplicate tx if multiple addresses of to the body share some common tx.

We also added `fromTs` and `toTs` optional query params to that endpoint to ease the pagination work.

### DIFF
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -5472,6 +5472,28 @@
         "parameters": [
           {
             "schema": {
+              "format": "int64",
+              "type": "integer",
+              "minimum": "0"
+            },
+            "in": "query",
+            "name": "fromTs",
+            "description": "inclusive",
+            "required": false
+          },
+          {
+            "schema": {
+              "format": "int64",
+              "type": "integer",
+              "minimum": "0"
+            },
+            "in": "query",
+            "name": "toTs",
+            "description": "exclusive",
+            "required": false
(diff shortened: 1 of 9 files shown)


h0ngcha0 (approved):
LGTM :+1:

polarker:
Why this method is `POST` instead of `GET`?

tdroxler:
> Why this method is `POST` instead of `GET`?

You can query transactions for multiple addresses, you pass them in the POST body. As we paginate we don't return more transactionos, limit is the same as the `GET` for one address.

polarker:
> > Why this method is `POST` instead of `GET`?
> 
> You can query transactions for multiple addresses, you pass them in the POST body. As we paginate we don't return more transactionos, limit is the same as the `GET` for one address.

POST is usually used to create a resource
```
POST is used to send data to a server to create/update a resource.
```

tdroxler:
> POST is usually used to create a resource
> 
> ```
> POST is used to send data to a server to create/update a resource.
> ```
yes, but REST doesn't have a nice way to request multiple items, so it's quite common to use POST for that kind of situation. The other solution is to go with chain of query param: `?address=<addr1>&address=<addr2>`, but you have a risk to reach the url length limit, especially with addresses that can be long.

polarker:
I keep forgetting that `GET` does not allow data to be sent to the server. We don't have other choices then.

polarker (approved):

PR #570, opened 2024-10-08
Expose max and default pagination limit in openapi

tdroxler:
We also revet the previous value to allow application to migrate to new values.

The idea is to already expose the future values, so FE can start using non-hardcoded value, once we'll have new releases and most users have migrated, we could reduce the max limit back to 20.


polarker (approved):
LGTM

PR #573, opened 2024-10-12
add: allow generic prometheus source

sven-hash:
If an existing grafana instance is used to import this dashboard, the prometheus source id will not be the same. With the proposal below you can specify the source you want to use


h0ngcha0 (approved):
LGTM :+1:

PR #574, opened 2024-10-23
Respect the `directCliqueAccess` flag when fetching blocks and events

Lbqds:


polarker:
@tdroxler It's unbelievable that `directCliqueAccess` was not implemented properly until today.

Issue #557, opened 2024-07-03
Integrate data from Mobula.io for price related endpoints


nop33:
As discussed on Slack, we want to start using Mobula as our primary source of ALPH and token prices, with Coingecko being the backup option. We want to use the explorer backend as a proxy to avoid being rate-limited.

Related issue:
- https://github.com/alephium/alephium-frontend/issues/693

Related resources:
- https://docs.mobula.io/

mvaivre:
We have multiple projects pushing to get their token price visible in our wallets cc @MaudSimon 

This would be great to start looking into this soon :)

PR #576, opened 2024-10-30
Mobula integration

tdroxler:
resolves https://github.com/alephium/explorer-backend/issues/557

Thx to Mobula, we can now find the price for almost all our token list, only ~40 tokens don't have a price.

In this PR we are now taking the token list and grab the price for all of them from mobula.

As previously we cache the data and reload every 5min.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/MarketService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/MarketService.scala
@@ -37,7 +37,7 @@ import org.alephium.util.{discard, Duration, Hex, Math, Service, TimeStamp}
 trait MarketService {
   def getPrices(
       ids: ArraySeq[String],
-      chartIurrency: String
+      chartCurrency: String
   ): Either[String, ArraySeq[Option[Double]]]
 
   def getExchangeRates(): Either[String, ArraySeq[ExchangeRate]]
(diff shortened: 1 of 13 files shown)


app/src/main/scala/org/alephium/explorer/service/MarketService.scala
```diff
@@ -30,46 +30,50 @@ import org.alephium.explorer.cache._
...
-import org.alephium.util.{Duration, Math, Service, TimeStamp}
+import org.alephium.protocol.Hash
+import org.alephium.protocol.model.{Address, ContractId}
+import org.alephium.util.{discard, Duration, Hex, Math, Service, TimeStamp}
 
 trait MarketService {
   def getPrices(
       ids: ArraySeq[String],
-      currency: String
+      chartIurrency: String
```
h0ngcha0, line 40: [Nit] `chartCurrency` ?

app/src/main/scala/org/alephium/explorer/service/MarketService.scala
```diff
@@ -102,7 +106,7 @@ object MarketService extends StrictLogging {
      * We use our `AsyncReloadCache` that always return the latest cached value
      * even if it's expired, like this we guarantee to always return fast a data.
      * We use an `Either` value because we cannot control what's returned by
-     * coingecko and it might be that their endoints return something else.
+     * coingecko and it might be that their endoints ret60rn something else.
```
h0ngcha0, line 109: [Nit] Is this intentional?

h0ngcha0 (commented):
LGTM, only minor comments.

app/src/main/scala/org/alephium/explorer/service/MarketService.scala
tdroxler: oops definitively not

PR #577, opened 2024-10-31
WIP - Fix flaky test

tdroxler:


app/src/test/scala/org/alephium/explorer/GenCoreApi.scala
```diff
@@ -271,7 +272,7 @@ object GenCoreApi {
         height = height,
         chainFrom = previousBlocks.last.chainFrom,
         chainTo = previousBlocks.last.chainTo,
-        transactions = AVector.from(txs),
+        transactions = AVector.from(txs :+ coinbaseTx),
```
tdroxler, line 275: the generated data was also leading to some flakyness, as the last tx in a block was not a coinbase one.

h0ngcha0 (approved):
LGTM :+1: 

As discussed, try it on testnet and monitor the CPU & memory usage

If it works well, will run on mainnet. If we need to stop the migration we will revert to the previous version.

Issue #560, opened 2024-08-22
Sub-contracts endpoint has duplicates


tdroxler:
![image](https://github.com/user-attachments/assets/815d18e5-129e-4716-9961-20187ef2b570)

PR #578, opened 2024-11-06
Fix sub-contracts endpoint returning duplicate

tdroxler:
Fixes: https://github.com/alephium/explorer-backend/issues/560

In database we can have multiple entry for a same contract, if it was destroyed and recreated, that's why we had duplicated values


h0ngcha0 (approved):
LGTM :+1:

PR #579, opened 2024-11-14
Remove unnecessary `DISTINCT`

tdroxler:
Now that we fixed the coinbase value and as we are querying for only 1 address, we will already receive a distinct set of transactions.


h0ngcha0 (approved):
LGTM :+1:

PR #583, opened 2024-11-25
[WIP] Fix chart price conversion

tdroxler:
while migrating to mobula, the base currency value was changed to `usd` as it's the one used by mobula, but then we didn't use anymore this value and we were requesting the chart price from coingecko in `usd`, but we were expecting charts to be in `btc`.

Before the fix the alph price in usd:

image

after the fix 

image

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/MarketService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/MarketService.scala
@@ -50,7 +50,6 @@ trait MarketService {
 }
 
 object MarketService extends StrictLogging {
-  val baseCurrency: String = "usd"
 
   object Impl {
     def default(marketConfig: ExplorerConfig.Market)(implicit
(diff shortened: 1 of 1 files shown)


app/src/main/scala/org/alephium/explorer/service/MarketService.scala
```diff
@@ -212,12 +212,14 @@ object MarketService extends StrictLogging {
         assets: ArraySeq[TokenList.Entry],
         retried: Int
     ): Future[Either[String, ArraySeq[Price]]] = {
-      handleResponseAndRetryOnTooManyRequests(
+      handleResponseAndRetryWithCondition(
         "mobula/price",
         response,
+        _.code != StatusCode.Ok,
```
h0ngcha0, line 218: Q: So in case of too many requests, they return 403 instead? 
Maybe we should not retry for 4xx error codes since those are usually non-transient. WDYT?

tdroxler: yes they return 403, but it's actually transient, as if you retry few seconds latter it's all good.

PR #585, opened 2024-12-03
Improve balance query

tdroxler:
The balance query was slow on addresses with a lot of unspent output.

Reducing the scope of the query with the address and the finalized timestamp + specific indexes improves a lot the query.

Here is some benchmarks, I made a lot of them but make a resume here

Example are done with some of our slowest addresses (extracted from the google cloud logs), they are in the top addresses in term of unspent output, around 150'000 of them. Queries were done on my local explorer backend. 

We have those various scenario:

1. Current state
2. With indexes
3. With indexes + new query

On the first query we see 1,2,3

image

On two next images, we see the query plan for 2 and 3 on two similar addresses
I was always shutting down pg and clearing caches to have a cold start.
We can see both new indexes being used and in the second case, the filtering on the indexes.

image
image

Next image is 1 and 2

We can see the query plan of 1 being more complicated
image

 I also made a sanity check, took around 5K random addresses and compared my new endpoint with the production one, everything's fine.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
@@ -92,8 +92,9 @@ object OutputSchema extends SchemaMainChain[OutputEntity]("outputs") {
 
   def createNonSpentOutputCoveringIndex(): DBActionW[Int] =
     sqlu"""
-      CREATE INDEX IF NOT EXISTS non_spent_output_covering_idx
-      ON #${name} (address, main_chain, spent_finalized, key, amount, lock_time)
+      CREATE INDEX non_spent_output_covering_include_idx
+      ON #${name} (address, main_chain, spent_finalized, key)
+      INCLUDE (amount, lock_time)
       WHERE spent_finalized IS NULL AND main_chain = true;
     """
   val table: TableQuery[Outputs] = TableQuery[Outputs]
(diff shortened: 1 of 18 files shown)


app/src/main/scala/org/alephium/explorer/persistence/schema/OutputSchema.scala
```diff
@@ -90,5 +90,11 @@ object OutputSchema extends SchemaMainChain[OutputEntity]("outputs") {
   def createNonSpentIndex(): DBActionW[Int] =
     sqlu"create unique index if not exists non_spent_output_idx on #${name} (address, main_chain, key, block_hash) where spent_finalized IS NULL;"
 
+  def createNonSpentOutputCoveringIndex(): DBActionW[Int] =
```
h0ngcha0, line 93: Do we need to use the INCLUDE keyword to create this covering index? Would be great to run a `EXPLAIN ANALYZE` for the balance query to check if we are doing index only scan.

tdroxler: as discussed privately. Using `INCLUDE` provide same performance, but reduce the index size: 
```
explorer=# SELECT 
    pg_size_pretty(pg_relation_size('non_spent_output_covering_idx')) AS index_size;
 index_size 
------------
 1085 MB
(1 row)

explorer=# SELECT 
    pg_size_pretty(pg_relation_size('non_spent_output_covering_include_idx')) AS index_size;
 index_size 
------------
 690 MB
(1 row)
```

So I just pushed a commit to use that.

h0ngcha0 (approved):
LGTM :+1:

PR #589, opened 2024-12-14
Cleanup Token URI with Null Bytes

h0ngcha0:


app/src/test/scala/org/alephium/explorer/persistence/queries/TokenQueriesSpec.scala
```diff
@@ -143,6 +145,18 @@ class TokenQueriesSpec extends AlephiumFutureSpec with DatabaseFixtureForEach wi
       result is tokens
     }
 
+    "insert and list nft metadata with token uri that contains null bytes" in {
+      val metadata = nftMetadataGen.sample.get
+      val tokenUri = "8e6877d8dee708506625a77d73f62210dce24564cfa9ec93c0df02f2ba519b00"
```
h0ngcha0, line 150: This is the problematic URI

PR #593, opened 2024-12-19
Improve caching of number of blocks

tdroxler:
Instead of doing a count in PG, which is a very slow process, we rather rely on the `latest_blocks` table, we can just sum up all height of the latest inserted block.


polarker (approved):
LGTM

PR #598, opened 2025-01-09
Optimize Finalizer Service

tdroxler:
Instead of a big join, we search first txs that need to be finalized, find the corresponding inputs, which can be done quickly with the tx hash and block hash as we have an index for that. The corresponding outputs are then updated.

Still need to fix the `TokenSupplyServicSpec`


app/src/main/scala/org/alephium/explorer/persistence/Migrations.scala
```diff
@@ -91,11 +91,33 @@ object Migrations extends StrictLogging {
    */
   def migration4: DBActionAll[Unit] = DBIOAction.successful(())
 
+  def migration5(implicit ec: ExecutionContext): DBActionAll[Unit] = {
```
h0ngcha0, line 94: Lets remove this as we discussed.

Issue #590, opened 2024-12-16
store metadata as byte


tdroxler:
as we don't control what stored inside the `ByteVec`, like the token-uri, we can have invalid data like in [this case](https://github.com/alephium/explorer-backend/pull/589).

We need to store those data, like the token uri as bytea in PG and do the string conversion at api level.

tdroxler:
Concerned tables:

* `nft_metadata`
* `fungible_token_metadata`
* `nft_collection_metadata`

PR #602, opened 2025-01-20
Quifix for null byte char in string

tdroxler:
testnet is down due to this issue https://github.com/alephium/explorer-backend/issues/590
We retake the same quickfix as last time, but apply it everywhere, as implementing the above issue might take a bit more time


h0ngcha0 (commented):
Could you please build an image? I will apply it on testnet and see if it fixes the issue.

h0ngcha0 (approved):
LGTM. Confirmed that this fixed the issue on testnet.

Issue #481, opened 2023-07-06
Check if raw SQL is used everywhere


tdroxler:
We know by experience that slick dsl is super slow, maybe we still use it in some places.
The goal of this issue is to scan all our code base and find those last piece of code, hopefully there isn't anymore.

tdroxler:
I found some in: 

* AppStateQueries
* HashrateService
*  BlockQueries
* TransactionQueries

also some `insertOrUpdate` elsewhere but for those we have we can stay with slick

PR #604, opened 2025-01-24
Various low hanging

tdroxler:
Regarding: Prevent `SttpBackend` to throw error on close
    
    When stopping the `MarketService`, we have to close the `SttpBackend`
    unfortunately we don't have much control over it and when it's closed
    while doing a request, it's throwing an error and we were logging that
    error.
    
    It's correct in a way, but it was confusing on the reason why the
    explorer-backend was stopping.
    
    For example if the node stop while the market service is updating its
    caches, an error was logged, saying the caches couldn't get updated.
    
    This was confusing and let user think the market service was the reason
    for the explorer-backend to stop.
    
    We are now making sure that if we catch an error while the service is
    stopped or not running, we don't throw the error.
    
    On the other hand if the service is running, we do want to log the
    error.

Resolves: https://github.com/alephium/explorer-backend/issues/481

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/HashrateService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/HashrateService.scala
@@ -101,6 +101,7 @@ case object HashrateService extends StrictLogging {
       FROM hashrates
       WHERE interval_type = $intervalType
       ORDER BY block_timestamp DESC
+      LIMIT 1
     """.asAS[TimeStamp].headOrNone
 
   private def findLatestHashrateAndStepBack(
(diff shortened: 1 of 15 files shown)


app/src/main/scala/org/alephium/explorer/service/HashrateService.scala
```diff
@@ -93,19 +93,22 @@ case object HashrateService extends StrictLogging {
...
-      .headOption
+  private def findLatestHashrate(intervalType: IntervalType)(implicit
+      ec: ExecutionContext
+  ): DBActionR[Option[TimeStamp]] =
+    sql"""
+      SELECT block_timestamp
+      FROM hashrates
+      WHERE interval_type = $intervalType
+      ORDER BY block_timestamp DESC
+    """.asAS[TimeStamp].headOrNone
```
simerplaha, line 104: Question: Does this query fetch multiple rows from the DB and return only the first row (`headOption`) after reading multiple rows? Would adding `LIMIT 1` to the query be helpful here?

simerplaha (approved):
Just one question. Otherwise, LGTM.

app/src/main/scala/org/alephium/explorer/service/HashrateService.scala
tdroxler: good point, indeed that's better so we don't grab all data from DB just to keep only the first value.

I'll make a quick check in all our codebase, we might have similar places

PR #609, opened 2025-03-11
Add non-coinbase txs count in the history endpoint

tdroxler:
Those values are nice to have in order to see the real sent transactions over the coinbase ones.

Not really sure about the all `nonCoinbase` naming, but it's at least very explicit.

No extra queries are needed and we keep the same performance of the previous optimization. 

We'll just need to decide if we want to reset the past month of history to re trigger the computation and have the non-coinbase count.

image

### DIFF
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -6689,17 +6689,19 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/TimedCount"
+                    "$ref": "#/components/schemas/TimedTxCount"
                   }
                 },
                 "example": [
                   {
                     "timestamp": 1611041396892,
-                    "totalCountAllChains": 10000000
+                    "totalCountAllChains": 10000000,
+                    "nonCoinbaseCountAllChains": 800
                   },
                   {
                     "timestamp": 1611041396892,
-                    "totalCountAllChains": 10000000
+                    "totalCountAllChains": 10000000,
+                    "nonCoinbaseCountAllChains": 800
                   }
                 ]
               }
@@ -6833,7 +6835,7 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/PerChainTimedCount"
+                    "$ref": "#/components/schemas/PerChainTimedTxCount"
                   }
                 },
                 "example": [
@@ -6843,12 +6845,14 @@
                       {
                         "chainFrom": 1,
                         "chainTo": 2,
-                        "count": 10000000
+                        "count": 10000000,
+                        "nonCoinbaseCount": 800
                       },
                       {
                         "chainFrom": 1,
                         "chainTo": 2,
-                        "count": 10000000
+                        "count": 10000000,
+                        "nonCoinbaseCount": 800
                       }
                     ]
                   },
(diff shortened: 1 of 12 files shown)


app/src/main/resources/explorer-backend-openapi.json
```diff
@@ -6689,17 +6689,19 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/TimedCount"
+                    "$ref": "#/components/schemas/TimedTxCount"
```
tdroxler, line 6692: `TimedCount` was originally meant to be a general case class that could be reused for any history data. But now that we introduce two count (total and non-coinbase), it's better to be explicit that it's for transactions count. It was only use there.

app/src/main/resources/explorer-backend-openapi.json
```diff
@@ -6833,7 +6835,7 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/PerChainTimedCount"
+                    "$ref": "#/components/schemas/PerChainTimedTxCount"
```
tdroxler, line 6838: [same as above](https://github.com/alephium/explorer-backend/pull/609/files#r1989402924)

app/src/main/scala/org/alephium/explorer/service/TransactionHistoryService.scala
```diff
@@ -166,79 +167,95 @@ case object TransactionHistoryService extends StrictLogging {
...
       SELECT block_timestamp, chain_from, chain_to, txs_count
       FROM block_headers
       WHERE block_timestamp >= $from
       AND block_timestamp <= $to
       AND main_chain = true
-    """.asAS[(TimeStamp, Int, Int, Long)]
+    """
+      .asAS[(TimeStamp, Int, Int, Long)]
+      .map(_.map { case (ts, from, to, value) =>
+        (ts, from, to, value, Some(value - 1))
```
tdroxler, line 222: That's were we get the `txsCount` for each block, as a block as only 1 coinbase tx, we can simply remove 1.

app/src/main/scala/org/alephium/explorer/service/TransactionHistoryService.scala
```diff
@@ -166,79 +167,95 @@ case object TransactionHistoryService extends StrictLogging {
...
+        val (count, nonCoinbaseCount) = data
+          .filter { case (timestamp, chainFrom, chainTo, _, _) =>
+            timestamp >= from &&
+            timestamp <= to &&
+            chainFrom == chainIndex.from.value &&
+            chainTo == chainIndex.to.value
+          }
+          .foldLeft((0L, 0L)) {
+            case ((currentCount, currentNonCoinbaseCount), (_, _, _, count, nonCoinbaseCount)) =>
+              (currentCount + count, currentNonCoinbaseCount + nonCoinbaseCount.getOrElse(0L))
```
tdroxler, line 185: we use `nonCoinbaseCount.getOrElse(0L)` as the DB type is Optional, as we introduce it later with a migration.
But during the processing here, we are guarantee to have a value, as four `hourly` the `txsCount` is fetch from the blocks table [here](https://github.com/alephium/explorer-backend/pull/609/files#r1989412943).
For `daily` we use the data stored in `hourly` so it will always have the freshly inserted data.

PR #610, opened 2025-03-19
Add API for mobula

tdroxler:
API key is optional, but if none is provided, the market service will return error when getting price.


h0ngcha0 (approved):
LGTM :+1:

PR #611, opened 2025-04-03
Add `Consensus` to config

tdroxler:
It was hardcoded as a quick fix for the Rhone update, but to anticipate Danube, it's finally moved as a config value


h0ngcha0 (approved):
LGTM :+1:

PR #613, opened 2025-04-10
Optimize transaction cache

tdroxler:
Previously in our `TransactionCache` every 10 minutes we were re-counting every txs in database, not optimal...

In this PR, we take advantage of the `FinalizerService`, which is grabbing every 10min every txs that needs to be finalized.
Here is the workflow:

* We keep in our `AppState` table the current count of finalized txs.
* When we finalize new txs, we add their count to the `finalized_tx_count` in `app_state`
* The `TransactionCache` is first reading the `finalized_tx_count` and only count txs that aren't finalized.

With this solution, whatever the number of transactions, the cache will only count a very small set of transactions.


h0ngcha0 (approved):
LGTM :+1:

PR #614, opened 2025-04-14
Define config files base on the network

tdroxler:
As done in the node, there are some values like the hard-fork timestamps that are different base on the network.


app/src/test/scala/org/alephium/explorer/config/ExplorerConfigSpec.scala
```diff
@@ -36,8 +36,13 @@ class ExplorerConfigSpec extends AlephiumSpec with ScalaCheckDrivenPropertyCheck
...
   "ficus" should {
     "load config" in {
-      val typesafeConfig = ConfigFactory.load()
-      Try(typesafeConfig.as[ExplorerConfig]("alephium")) is a[Success[ExplorerConfig]]
+      // We make sure every config file is valid
+      Seq(NetworkId.AlephiumMainNet, NetworkId.AlephiumTestNet, NetworkId.AlephiumDevNet).foreach {
+        networkId =>
+          val typesafeConfig =
+            ConfigFactory.parseResources(s"application-${networkId.networkType}.conf").resolve()
+          Try(typesafeConfig.as[ExplorerConfig]("alephium")) is a[Success[ExplorerConfig]]
```
h0ngcha0, line 44: [Nit] Shall we verify the value is correctly loaded for different network id?

h0ngcha0 (approved):
LGTM :+1:

app/src/test/scala/org/alephium/explorer/config/ExplorerConfigSpec.scala
tdroxler: good idea, I'll add it to the test

PR #616, opened 2025-05-01
Sync with WebSocket

tdroxler:
Update of https://github.com/alephium/explorer-backend/pull/478 using the new websocket of full node and the `WsClient`
This PR needs the latest version of [that node PR](https://github.com/alephium/dev-alephium/pull/1099)

the strategy stay the same:

* Sync with `BlockFlowSyncService` as usual
* Once we are up-to-date: latest block is less than 30s old we move to
  websocket syncing
* If some blocks are missed between the last sync and the opening of the websocket, it will eventually recover them
  as we always make sure the parents are downloaded before inserting new blocks.
* If websocket messages are late: > 30s old, we close the websocket and
  move back to http syncing. This can happen in case of network issue,
  this is safer as we could suddenly receive lots of message through the
  websocket and the DB can't follow (happen once in my tests)
* If the websocket close for anyreason we also move back to http
  syncing.

The main idea is to always rely and fall back on our BlockFlowSyncService which is well tested in production.

I tried various edge cases, like cutting my network etc.

Still require some testing and extracting some config values

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/ExplorerState.scala
+++ b/app/src/main/scala/org/alephium/explorer/ExplorerState.scala
@@ -25,6 +25,7 @@ import slick.jdbc.PostgresProfile
 
 import org.alephium.explorer.cache.{BlockCache, MetricCache, TransactionCache}
 import org.alephium.explorer.config.{BootMode, ExplorerConfig}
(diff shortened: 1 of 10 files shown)


tdroxler:
another branch is working on that.

PR #617, opened 2025-05-20
Add indexes for groupless address columns

tdroxler:
A migration is needed to drop a index that has a wrong name + it was indexing the `address` field too while it only need `groupless_address`

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/persistence/schema/TokenTxPerAddressSchema.scala
+++ b/app/src/main/scala/org/alephium/explorer/persistence/schema/TokenTxPerAddressSchema.scala
@@ -62,10 +62,10 @@ object TokenPerAddressSchema
 
   def createConcurrentIndexes()(implicit ec: ExecutionContext): DBActionW[Unit] =
     for {
-      _ <- createGrouplessAddressTimestampIndex()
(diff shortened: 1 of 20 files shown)


app/src/main/scala/org/alephium/explorer/persistence/schema/TokenTxPerAddressSchema.scala
```diff
@@ -57,5 +60,17 @@ object TokenPerAddressSchema
       CommonIndex.timestampIndex(this)
     )
 
+  def createConcurrentIndexes()(implicit ec: ExecutionContext): DBActionW[Unit] =
+    for {
+      _ <- createGrouplessAddressTimestampIndex()
+    } yield ()
+
+  def createGrouplessAddressTimestampIndex(): DBActionW[Int] =
```
h0ngcha0, line 68: [Nit] Is the naming of the function correct? Seems that it creates index for groupless address and tokens.

tdroxler: yep....copy-pasting is baaaad.... :sweat_smile:

PR #619, opened 2025-06-12
Bring back coingecko prices

tdroxler:
We now try to fetch prices from both mobula and coingecko, in case one of them fail, we then merge the prices.

We are currently favoring mobula prices are the are more accurate, but since we lately had issue querying them, coingecko is here as a fallback.

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/service/MarketService.scala
+++ b/app/src/main/scala/org/alephium/explorer/service/MarketService.scala
@@ -123,13 +123,13 @@ object MarketService extends StrictLogging {
      */
     private val mobulaPricesCache: AsyncReloadingCache[Either[String, ArraySeq[Price]]] =
       AsyncReloadingCache[Either[String, ArraySeq[Price]]](
-        Left("Price data not fetched"),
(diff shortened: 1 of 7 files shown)


app/src/main/scala/org/alephium/explorer/service/MarketService.scala
```diff
@@ -115,11 +121,17 @@ object MarketService extends StrictLogging {
      * We use an `Either` value because we cannot control what's returned by
      * coingecko and it might be that their endoints return something else.
      */
-    private val pricesCache: AsyncReloadingCache[Either[String, ArraySeq[Price]]] =
+    private val mobulaPricesCache: AsyncReloadingCache[Either[String, ArraySeq[Price]]] =
+      AsyncReloadingCache[Either[String, ArraySeq[Price]]](
+        Left("Price data not fetched"),
```
h0ngcha0, line 126: [Nit] Probably `Price data not fetched from Modula`, same for Coingecko

h0ngcha0 (approved):
LGTM

PR #620, opened 2025-06-12
Run `MarketService` even without mobula api-key

tdroxler:
Now that coingecko act as fall back, without an api-key we can still get coingecko results.


h0ngcha0 (commented):
LGTM

PR #626, opened 2025-06-27
Flaky db test

tdroxler:
The background indexes were introduces lately to avoid down time when deploying a new version, in case the indexes were long to create, but it was making the `DatabaseSpec` very flaky. 

This new option actually also helped to find a missing `IF EXISTS` as it was running in background until then, it was not spotted by tests.


h0ngcha0 (approved):
LGTM

PR #635, opened 2025-07-16
Load correct default values from env variable

tdroxler:
The default value for mainnet or testnet was selected only based on the network id found in `EXPLORER-HOME/user.conf` and not the `BLOCKFLOW_NETWORK_ID`.

Here two screenshot to show the correct file choosen base on the env variable, once with sbt, the other when using the jar file:

image
image


app/src/main/scala/org/alephium/explorer/config/ExplorerConfig.scala
```diff
@@ -66,12 +66,23 @@ object ExplorerConfig {
...
+   * Otherwise try to use the env variable `BLOCKFLOW_NETWORK_ID`.
+   * Otherwise, default to AlephiumMainNet.
+   */
   def parseNetworkId(config: Config): Try[NetworkId] = {
     val keyPath = "alephium.blockflow.network-id"
     if (config.hasPath(keyPath)) {
       Try(config.as[NetworkId](keyPath))
     } else {
-      Success(NetworkId.AlephiumMainNet)
+      sys.env.get("BLOCKFLOW_NETWORK_ID").flatMap(_.toIntOption).flatMap { id =>
```
h0ngcha0, line 79: Should `BLOCK_NETWORK_ID` has priority if set? It's like [this](https://github.com/alephium/alephium/blob/master/flow/src/main/scala/org/alephium/flow/setting/Configs.scala#L101-L114) in full node.

PR #639, opened 2025-08-01
Remove `UNIQUE` constraint of some index

tdroxler:
Two entries for those indexes could occurs if `addr:0` is sending for example to `addr:1`.

One input will contain the address from group 0, while one output the address from group 1.

Both input and output will add an entry in the `transaction_per_addresses`, which is fine as it's two different addresses, but for indexes, as we want to index only the `groupless_address`, as they have the same one, it returns a uniqueness error.

The `UNIQUE` was wrongly used, thinking in those cases it was only creating one entry in  the index for duplicate values.

This commit remove the indexes, but replace them with the same index without the uniqueness constraint.

Performance will still be good even with some duplicates and the uniqueness when querying transactions for groupless_address is already handled [here](https://github.com/alephium/explorer-backend/blob/7562e9bedc1bce836919b9a25b59349e2df7bad4/app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala#L159) with a `DISTINCT`


h0ngcha0 (approved):
LGTM

PR #643, opened 2025-08-26
Conflicted txs

tdroxler:
## Overview

This pull request implements robust handling of the new `conflictedTxs` information from the node, enabling the explorer to correctly reflect transaction conflicts arising from parallel block mining. To improve block times, our protocol allows two blocks to be mined in parallel, potentially spending the same `output_ref`. Once the intra-block is mined, it decides which transaction is accepted and flags the conflicting one as `conflicted`.

## Conflict Detection Logic

When syncing, upon receiving an intra-block, we need to revisit recent blocks to detect and resolve any transaction conflicts. There are two main scenarios:

1. **Duplicate Usage of an Output Reference**
   - Find all `(block_hash, tx_hash)` pairs where the same `output_ref` is spent by two transactions on the `main_chain`.

2. **Output Reference Originates from a Conflicted Transaction**
   - Find all `(block_hash, tx_hash)` pairs where the `output_ref_tx_hash` is on the `main_chain` and marked as `conflicted`.

For each detected conflict, we query the node for the updated `conflicted_txs` value and update our records accordingly.

## Handling Re-orgs

If two blocks compete for a transaction but, after a re-org, only one remains on the `main_chain`, we must:
- Identify duplicate `output_ref` entries marked as `conflicted`, where one is on the `main_chain` and the other is not.
- Update the conflicted status by querying the node as usual.

## Finalization

During data finalization, we perform a comprehensive check of all blocks flagged as `conflicted` to ensure accuracy, since these records are no longer modified once finalized.

## Querying Data

With the introduction of `conflictedTxs`, we now need to handle these cases carefully when querying data:

- **Block Transactions:**  
  When retrieving all transactions for a block, we must include both regular and conflicted transactions to accurately reflect the node’s data.

- **Address Transactions and Balances:**  
  When requesting the transaction history or balance for an address, conflicted transactions should be excluded. This ensures users see only valid (non-conflicted) transactions in their wallet or history, matching the canonical chain state.

This distinction is important to maintain consistency between explorer results and node responses, and to provide users with correct transaction and balance information.

### DIFF
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -8586,6 +8586,13 @@
             "items": {
               "$ref": "#/components/schemas/GhostUncle"
             }
+          },
+          "conflictedTxs": {
+            "type": "array",
+            "items": {
+              "type": "string",
+              "format": "32-byte-hash"
+            }
           }
         }
       },
--- a/app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/EndpointExamples.scala
@@ -137,7 +137,8 @@ object EndpointExamples extends EndpointsExamples {
       hashRate = HashRate.a128EhPerSecond.value,
       parent = Some(blockHash),
       mainChain = true,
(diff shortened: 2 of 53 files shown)


Lbqds (approved):
I didn’t check all the details, but the approach for handling conflicted txs LGTM 👍

PR #644, opened 2025-08-27
Fix Address Transaction Count Cache for `ReadOnly` Mode

tdroxler:
Previously, the address transaction count cache was optimized by updating the cached value "on-demand"—whenever a count was requested, the new value was written back to the database. However, this approach caused failures when running in `ReadOnly` mode, since database writes are not permitted.

This commit introduces separate cache implementations for read-only and write modes:
- **ReadOnly mode:** Uses an in-memory cache. Initial values are loaded from the database if available, but subsequent updates remain only in memory to avoid write operations.
- **Write mode:** Continues to use the database-backed cache, supporting persistent updates as before.

This change ensures correct and efficient behavior for both read-only and write-enabled instances.


h0ngcha0 (approved):
LGTM :+1:

PR #646, opened 2025-09-09
Fix incorrect field order in SQL queries due to use of SELECT *

tdroxler:
We were using `SELECT *` in several queries and mapping the results directly to entity models. This approach is fragile and led to issues on testnet/mainnet.

For example, when we added the `conflicted` field to the transaction entity (positioned below `main_chain` in the model), a fresh database created the column in the expected order. However, in testnet/mainnet, where `conflicted` was added via migration, the column was placed at the end. As a result, `SELECT *` returned columns in the wrong order, causing `CustomGetResult` to misalign fields.

This commit replaces `SELECT *` with explicit column selection to ensure correct field mapping regardless of column order in the database schema.

This was also not catch by our test as some models where not creating all data (see first commit).

Only the [getTransactionAction](https://github.com/alephium/explorer-backend/compare/fix-field-order-in-queries?expand=1#diff-4bff8d7d0a817ed4688ba4879cd265dc01e72c52b6f5fddf69654f094d72b1d6R109-R114) was throwing an error, but I took the opportunity to remove all our `SELECT *`.


h0ngcha0 (approved):
Nice catch. LGTM 👍

PR #648, opened 2025-09-11
Add `/transactions/decode-unsigned-tx` endpoint with inputs details

tdroxler:
Using the same format as the node, the only difference is that we can provide then input details.

Resolves https://github.com/alephium/explorer-backend/issues/642

### DIFF
--- a/app/src/main/scala/org/alephium/explorer/api/model/UnsignedTransaction.scala
+++ b/app/src/main/scala/org/alephium/explorer/api/model/UnsignedTransaction.scala
@@ -1,0 +1,33 @@
+// Copyright (c) Alephium
+// SPDX-License-Identifier: LGPL-3.0-only
+
+package org.alephium.explorer.api.model
+
+import scala.collection.immutable.ArraySeq
+
+import sttp.tapir.Schema
+
+import org.alephium.api.TapirSchemas._
+import org.alephium.explorer.api.Json._
+import org.alephium.json.Json._
+import org.alephium.protocol.model.TransactionId
+import org.alephium.util.U256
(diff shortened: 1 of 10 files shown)


app/src/main/scala/org/alephium/explorer/api/model/UnsignedTransaction.scala
```diff
@@ -0,0 +1,33 @@
...
+    hash: TransactionId,
+    version: Byte,
+    networkId: Byte,
+    scriptOpt: Option[String],
+    gasAmount: Int,
+    gasPrice: U256,
+    inputs: ArraySeq[Input],
+    outputs: ArraySeq[Output]
+) {
+  def chainFrom: Option[Int] =
```
h0ngcha0, line 26: Shall we return `Int` instead of `Option[Int]`? Since we should always have `chainFrom`.

h0ngcha0 (approved):
LGTM :+1:

app/src/main/scala/org/alephium/explorer/api/model/UnsignedTransaction.scala
tdroxler: well `chainFrom` is derived from the inputs [as in the node](https://github.com/alephium/alephium/blob/master/protocol/src/main/scala/org/alephium/protocol/model/UnsignedTransaction.scala#L58-L60), but as I don't control the data here, I thought it was safer to use `Option` in case inputs are empty.

But actually I realized that `chainFrom` isn't used, can't remember why it was introduced. I will simply remove it for now.

PR #651, opened 2026-01-15
Batch mobula request

tdroxler:
They have a limit of max 50 tokens, so we were not able to get price from them as our queries were failing.


h0ngcha0 (approved):
LGTM 👍

PR #659, opened 2026-03-25
Fix export transactions

tdroxler:
Previously, if a transaction export started but the user canceled it (or the connection dropped), the backend would just keep churning away in the background. This resulted in a ton of zombie DB calls and wasted CPU.

We saw this hit us in prod: EB was a bit slow, Nginx timed out the download, and the user (rightfully) clicked retry multiple times. This created a "stack" of simultaneous exports for the same request, which blew up our RAM.

**Changes**:

* We now listen for the Vert.x closeHandler. If the connection drops, we trigger a cancelToken.
* The internal export process is now reactive to this token and will kill the DB stream immediately.
* I changed the `batchSize` from 1 to 100. Previously, we were doing a DB round-trip for every single transaction, which was super inefficient (because I thought we wouldn't preserve transactions order with batch, which I was wrong). Now we fetch and process in chunks of 100, which significantly reduces the IO overhead and helps the stream keep up with Nginx timeouts.
* 
**Testing**:

I simulated the prod failure by setting a 1s Nginx timeout and [forcing a 2s delay between batches in EB](https://github.com/alephium/explorer-backend/blob/1d4d495cdaa6957668b4dfe6951b4e7f4d40b334/app/src/main/scala/org/alephium/explorer/service/TransactionService.scala#L493-L502). The internal process now correctly catches the disconnect and stops immediately.

### DIFF
--- a/app/src/main/resources/application.conf
+++ b/app/src/main/resources/application.conf
@@ -52,6 +52,10 @@ explorer {
     #How many transaction can be exported at max
     export-txs-number-threshold = 10000
 
+    #Number of batch while streaming
+    stream-batch-size = 100
+    stream-batch-size = ${?EXPLORER_STREAM_BATCH_SIZE}
+
     #Number of allowed parallelism when using `mapAsync` on streams.
     stream-parallelism = 8
     stream-parallelism = ${?EXPLORER_STREAM_PARALLELISM}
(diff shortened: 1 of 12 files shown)


h0ngcha0 (approved):
LGTM 👍

PR #663, opened 2026-06-10
Alph price

tdroxler:


app/src/main/resources/application.conf
```diff
@@ -95,6 +95,7 @@ explorer {
             # Others
             "WBTC"    = "wrapped-bitcoin",
             "WETH"    = "weth",
+            "WBNB"    = "wbnb"
```
nop33, line 98: I think these 2 don't really make sense. The `wETH` and `wBNB` on Alephium are not the same as the `wETH` and `wBNB` on Ethereum. The former are Alephium wrapped tokens of the ETH native token and the latter are Ethereum wrapped tokens of the ETH native token. Mapping the Alephium wrapped token to the price of the Ethereum wrapped token is as inaccurate as using the ETH token itself. Even weirder is mapping the Alephium wrapped BNB to the price of the Ethereum wrapped BNB. 

At the end of the day, is a negligible difference. In my opinion, the price for the Alephium wrapped tokens should come from our ecosystem (since it's wrapped on Alephium(, with fallback to the native assert (ETH/BNB) instead of the Ethereum wrapped version of it.

Does that make sense? 

In any case, it's a nit, since the price is pretty much the same. Just food for thought.

app/src/main/resources/application.conf
```diff
@@ -106,6 +107,10 @@ explorer {
         chart-symbol-name {
             "ALPH" = "alephium"
         }
+        # Symbols where CoinGecko price takes precedence over Mobula when both providers return a price.
+        coingecko-priority-symbols = [
+            "ALPH"
+        ]
```
nop33, line 113: 👍

nop33 (approved):

app/src/main/resources/application.conf
tdroxler: it does make sense, so it should rather comes from our powfi backend right? 
  
  We could keep it there in the mean time and switch to powfi when ready?

nop33: sounds good

PR #672, opened 2026-07-20
Add `POST /addresses/latest-transactions` endpoint

### LINKED ISSUE #669: Add batched POST /addresses/latest-transactions endpoint
## Request

A **batched** version of the existing per-address latest-transaction endpoint:

```
POST /addresses/latest-transactions
```

Body: a list of address hashes. Response: the latest transaction (hash + timestamp) for each.

This would be consistent with the batch endpoints that already exist (`POST /addresses/transactions`, `POST /addresses/used`, `POST /tokens`, `POST /tokens/fungible-metadata`).

## Why

Both Alephium wallets (desktop and mobile) drive their **entire** refresh cycle from one cheap sentinel query per address:

```
GET /addresses/{address}/latest-transaction
```

The wallet polls it, and when the returned hash changes it invalidates and refetches the heavy data (balances, token balances, tx lists). Every other query in the wallet is `staleTime: Infinity` and refetches *only* as a consequence of that signal. It is the single clock in the system.

The problem is that the endpoint is **per-address**, so the wallet issues **one HTTP request per address, per poll**:

- addresses used in the last 30 days: polled every **60s**
- dormant addresses: polled every **300s**

That gives us an unpleasant tradeoff. Polling faster means multiplying an already `O(N)` request count; keeping N low means slow updates. Concretely, today:

| Wallet | Requests/min at idle |
|---|---|
| 5 addresses | ~5 |
| 20 addresses | ~8 |
| 100 addresses | **~100** |

Wallets with many addresses are the worst hit, and the client self-throttles at 10 req/s, so a power user's refresh burst takes ~10 seconds just to drain.

## What batching buys

One request per poll regardless of address count. That lets us **poll much faster and still send fewer requests**:

| | latency (avg, end-to-end) | requests/min (100-addr wallet) |
|---|---|---|
| today, active address (60s) | ~41s | ~100 |
| today, dormant address (300s) | ~161s | (included above) |
| **batched, polled every 10s** | **~16s** | **~6** |

**Better on both axes at once**, which is rare. Dormant addresses - which is exactly where a fresh incoming payment tends to land - improve ~10x.

(The ~11s floor in those numbers is `~8s` block time + `~2.5s` because explorer-backend polls the node every 5s. The WebSocket-ingest work already in flight removes that second part for everyone.)

We evaluated server push (WebSockets/SSE) as the alternative. It would get this to ~11s - i.e. **a batched poll captures 84-97% of the achievable win at a tiny fraction of the cost.** That's why we're asking for this endpoint rather than a socket API.

## Suggested shape

Modelled on `POST /addresses/used`:

```
POST /addresses/latest-transactions
Content-Type: application/json

["addr1", "addr2", "addr3", ...]
```

```json
[
  { "address": "addr1", "hash": "…", "timestamp": 1752480000000 },
  { "address": "addr2", "hash": "…", "timestamp": 1752479000000 },
  { "address": "addr3" }
]
```

Notes / open questions:

- **Addresses with no transactions.** The current `GET` returns `404`, which the client already has to special-case. In a batch response it would be nicer to either omit them or return an entry with no `hash`/`timestamp`, rather than failing the whole request.
- **Max batch size.** Whatever is sane for you - we'll chunk client-side. Please document the limit and the behaviour on exceeding it. Something in the range of a few hundred would comfortably cover even large wallets in one request.
- **Ordering** does not matter to us; we key by address.
- We only need `hash` and `timestamp`. Anything more is wasted bytes on a request we intend to poll frequently.

## Client side

Frontend change is small and contained: it swaps the `refetchInterval` in `useAddressesDataPolling` and the whole invalidation cascade underneath is untouched. We're ready to implement as soon as the endpoint is available.

Ref: `alephium/alephium-frontend` - `packages/shared-react/src/features/dataPolling/useAddressesDataPolling.ts`, `packages/shared-react/src/api/queries/transactionQueries.ts`.

tdroxler:
Resolves https://github.com/alephium/explorer-backend/issues/669

Super fast even with our top 10 addresses by number of txs

```bash
 curltime -X 'POST' \                                                                                                                                                                                            15:51:01
  'http://localhost:9090/addresses/latest-transactions' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '[
   "1Gzp3Wc42fanrbAqknYmSaZLfC567bP3JWprDCiuJ8rbP",
 "1EbUwQWRfuXnEkvPTYxubp5cYsGo3foAMA9g9zvNQNSwW",
 "12jK2jHyyJTJyuRMRya7QJSojgVnb5yh4HVzNNw6BTBDF",
 "19WzSnmNC1SQ6v7RpFFXhpcMcFSiwM4nKTSdbwgSJfSHy",
 "14kSoi8pFMn2wMJFZJgGpFSA7GyNMtwWKnsKc1KidZ8Bu",
 "1PZonix2UoaguUfbbBnWevk4vod1m9MeJXBnkr7aqN76",
 "15AkQjovigzbQXGoHekLMb1prs1MTyRXFKSLw8VRy4quJ",
 "1LzhJdLG1SMMCPqRpnUJRs5wgLqRYcWZcRdG8baLeFbT",
 "1Ba2L3oUn6B9SR6DcwzyD1EK26JptfkoyXngKwG5s6XeP",
 "15AG4h7gy9EThb1riPwzzZh5v1yvPJwJ2ZaYieVJ4e1YE"
]
'

time_total: 0.107701
```

### DIFF
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -2554,6 +2554,139 @@
         }
       }
     },
+    "/addresses/latest-transactions": {
+      "post": {
+        "tags": [
+          "Addresses"
+        ],
+        "summary": "Get latest transaction information for given addresses",
+        "operationId": "postAddressesLatest-transactions",
+        "requestBody": {
+          "description": "List of addresses, max items: 80",
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "array",
+                "maxItems": 80,
+                "items": {
+                  "type": "string",
+                  "format": "address"
+                }
+              },
+              "example": [
+                "1AujpupFP4KWeZvqA7itsHY9cLJmx4qTzojVZrg8W9y"
+              ]
+            }
+          },
+          "required": false
+        },
+        "responses": {
+          "200": {
+            "description": "Ok",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/components/schemas/TransactionInfoPerAddress"
+                  }
+                },
+                "example": [
+                  {
+                    "address": "1AujpupFP4KWeZvqA7itsHY9cLJmx4qTzojVZrg8W9y",
+                    "transactionInfo": {
+                      "hash": "503bfb16230888af4924aa8f8250d7d348b862e267d75d3147f1998050b6da69",
+                      "blockHash": "bdaf9dc514ce7d34b6474b8ca10a3dfb93ba997cb9d5ff1ea724ebe2af48abe5",
+                      "timestamp": 1611041396892,
+                      "coinbase": false
+                    }
+                  }
+                ]
+              }
+            }
+          },
+          "400": {
+            "description": "BadRequest",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/BadRequest"
+                },
+                "example": {
+                  "detail": "Something bad in the request"
+                }
+              }
+            }
+          },
+          "401": {
+            "description": "Unauthorized",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Unauthorized"
+                },
+                "example": {
+                  "detail": "You shall not pass"
+                }
+              }
+            }
+          },
+          "404": {
+            "description": "NotFound",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/NotFound"
+                },
+                "example": {
+                  "resource": "wallet-name",
+                  "detail": "wallet-name not found"
+                }
+              }
+            }
+          },
+          "500": {
+            "description": "InternalServerError",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/InternalServerError"
+                },
+                "example": {
+                  "detail": "Ouch"
+                }
+              }
+            }
+          },
+          "503": {
+            "description": "ServiceUnavailable",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ServiceUnavailable"
+                },
+                "example": {
+                  "detail": "Self clique unsynced"
+                }
+              }
+            }
+          },
+          "504": {
+            "description": "GatewayTimeout",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/GatewayTimeout"
+                },
+                "example": {
+                  "detail": "The network is slow"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/addresses/{address}/mempool/transactions": {
       "get": {
         "tags": [
@@ -10178,6 +10311,23 @@
           },
           "coinbase": {
             "type": "boolean"
+          }
+        }
+      },
+      "TransactionInfoPerAddress": {
+        "title": "TransactionInfoPerAddress",
+        "type": "object",
+        "required": [
+          "address",
+          "transactionInfo"
+        ],
+        "properties": {
+          "address": {
+            "type": "string",
+            "format": "address"
+          },
+          "transactionInfo": {
+            "$ref": "#/components/schemas/TransactionInfo"
           }
         }
       },
--- a/app/src/main/resources/explorer-backend-openapi.json
+++ b/app/src/main/resources/explorer-backend-openapi.json
@@ -2560,6 +2560,7 @@
           "Addresses"
         ],
         "summary": "Get latest transaction information for given addresses",
+        "description": "Returns the latest transaction for each address. Duplicate addresses are deduplicated. Addresses with no transactions are omitted from the result.",
         "operationId": "postAddressesLatest-transactions",
         "requestBody": {
           "description": "List of addresses, max items: 80",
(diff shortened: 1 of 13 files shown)


app/src/main/resources/explorer-backend-openapi.json
nop33, line 1: This file is generated programmatically, right?

app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
```diff
@@ -91,6 +91,14 @@ trait AddressesEndpoints extends BaseEndpoint with QueryParams {
       .out(jsonBody[TransactionInfo])
       .summary("Get latest transaction information of a given address")
 
+  lazy val getLatestTransactionInfos
+      : BaseEndpoint[ArraySeq[ApiAddress], ArraySeq[TransactionInfoPerAddress]] =
+    baseAddressesEndpoint.post
+      .in(arrayBody[ApiAddress]("addresses", maxSizeAddresses))
+      .in("latest-transactions")
+      .out(jsonBody[ArraySeq[TransactionInfoPerAddress]])
+      .summary("Get latest transaction information for given addresses")
```
nop33, line 100: nit: Is this what is displayed on swagger? If that's the case, maybe it's worth explaining that the result will be keyed by address (so duplicated addresses collapse to unique) and that addresses with no txs are omitted. Does that make sense?

app/src/test/scala/org/alephium/explorer/persistence/queries/TransactionQueriesSpec.scala
```diff
@@ -210,6 +211,55 @@ class TransactionQueriesSpec
     txs should contain allElementsOf expected
   }
 
+  "get latest tx info by addresses" in new Fixture {
```
nop33, line 214: would it make sense to also make a test where an address with no txs is properly omitted?

app/src/test/scala/org/alephium/explorer/web/AddressServerSpec.scala
```diff
@@ -237,6 +247,34 @@ class AddressServerSpec()
     }
   }
 
+  "get latest transaction info for multiple addresses" should {
+    "return all different addresses info" in {
+      val addresses = ArraySeq.fill(3)(addressGen.sample.get)
```
nop33: will this generate 3 different addresses, or 3 identical?

app/src/main/scala/org/alephium/explorer/persistence/queries/TransactionQueries.scala
```diff
@@ -398,6 +398,43 @@ object TransactionQueries extends StrictLogging {
...
+             ORDER BY block_timestamp DESC, tx_order
+             LIMIT 1
+           """
+        }
+
+      val query =
+        s"""
+           SELECT lookup_address, address, ${TxByAddressQR.selectFields}
+           FROM (${branches.map(branch => s"($branch)").mkString(" UNION ALL ")}) matches
+           ORDER BY lookup_address
```
nop33: claude says this might not be necessary, but I can't confirm myself 😅 

<details><summary>Details</summary>

It's in the same TransactionDao.getLatestTransactionInfoByAddresses method, the two lines right after
  the query runs:
```
  run(getLatestTransactionInfoByAddressesAction(addresses).map { rows =>
    val infosByAddress = rows.map(row => row.lookupAddress -> row).toMap   // <- line 75
    addresses.flatMap { address =>                                         // <- line 76
      infosByAddress.get(address.toBase58).map { row => ... }
    }
  })
```
  Two steps, and together they make the SQL ordering irrelevant:

  1. Line 75 takes the ordered rows and immediately turns them into infosByAddress, a Map. A Map is a
  hash table with no meaningful iteration order, so whatever order the SQL returned is thrown away the
  moment .toMap runs.
  2. Line 76 builds the output by walking the original addresses (addresses.flatMap), not the rows. So
  the final sequence comes out in request order (deduped, minus the omitted empties), regardless of how
  the query sorted.

  So the query's ORDER BY lookup_address never reaches the response. The output order is decided
  entirely by the addresses.flatMap on line 76.

  An inner ORDER BY could matter if
  it influenced which row wins on a duplicate key in .toMap (Scala keeps the last pair on a key
  collision). But it can't here, because each address branch has LIMIT 1 and the server already deduped
  with .distinct, so every lookupAddress appears at most once. No collisions, no tie to break, so the
  sort is genuinely dead code. 

</details>

nop33 (approved):
Just nit comments, LGTM!

I tested this with your rc, here's the draft PR (I'll wait for a web3 rc before marking it ready)

- https://github.com/alephium/alephium-frontend/pull/1683

app/src/main/resources/explorer-backend-openapi.json
tdroxler: yes: `sbt tools/run`

app/src/main/scala/org/alephium/explorer/api/AddressesEndpoints.scala
tdroxler: image
this is one is a summary, so just a onle line short explanation. I can add a `description` explaining more details

app/src/test/scala/org/alephium/explorer/web/AddressServerSpec.scala
tdroxler: different, but  I’ll add a check to guarantee the uniqueness

PR #676, opened 2026-08-12
Fix contract update

tdroxler:
When we catch up  the syncing, we sync 30min of block at a time, but if suddenly there are a lot of contracts event within those 30min, it was crashing the app as the DB couldn't update in parrallel every contracts.

This commit fix it by applying the same [foldFutures as for tokens](https://github.com/alephium/explorer-backend/blob/612434c14cf99ba8ff655d1ede052e88de1251e0/app/src/main/scala/org/alephium/explorer/service/TokenService.scala#L276)

The first commit show the issue, second commit fix it


h0ngcha0 (approved):
LGTM 👍

PR #679, opened 2026-09-23
Use Powfi Prices For Selected Alephium Ecosystem Tokens

h0ngcha0:


app/src/main/scala/org/alephium/explorer/service/market/MarketService.scala
```diff
@@ -230,12 +243,93 @@ object MarketService extends StrictLogging {
...
+      for {
+        token0 <- tokenField(state.immFields, token0Index, tokens)
+        token1 <- tokenField(state.immFields, token0Index + 1, tokens)
+        raw    <- rawPrice
+        price0 = raw * math.pow(10, (token0.decimals - token1.decimals).toDouble)
+        price <-
+          if (token0.symbol == symbol) {
+            Right(PowfiPrice(symbol, token1.symbol, price0))
+          } else if (token1.symbol == symbol) {
+            Right(PowfiPrice(symbol, token0.symbol, 1 / price0))
```
polarker: In the extreme case, if price0 becomes zero due to zero liquidity, will this terminate the backend or some services due to exceptions?

h0ngcha0: This token will get no price instead. `price0` is a Double, `1 / 0.0` gives `Infinity` instead of throwing. The next line `(Either.cond(price.price > 0 && !price.price.isInfinite, …))` rejects `Infinity` so that token just gets no price.

Added a test in the latest commit.

polarker (approved):

